PR #25, opened 2024-07-27
fixes a bug which turns a remote_handle64 into a undefined struct

alexander-pick:
This bugfix fixes an issue which turns a remote_handle64 into a struct with undefined content. If you just cast the pointer and refer a struct field later it will be offset to the pointer target and not point to local which is the third member in the struct. As the struct is used later in the function this will lead to undefined behavior or error states.


src/fastrpc_apps_user.c
```diff
@@ -693,7 +693,9 @@ static inline bool is_valid_local_handle(int domain, struct handle_info *hinfo)
 }
 
 static int verify_local_handle(int domain, remote_handle64 local) {
-  struct handle_info *hinfo = (struct handle_info *)(uintptr_t)local;
+  struct handle_info *hinfo = malloc(sizeof(struct handle_info));
```
ekanshibu, line 696: you can "free(hinfo)" before returning nErr.

ekanshibu: also, how will other entries of hinfo(hlist, remote) get populated with this change, I'm trying out container_of() approach for this.

alexander-pick: I looked into this a bit more in depth today and saw that the caller of this functions (2 at this time) do similar things. One is easy to fix but the second one (get_domain_from_handle()) might be an issue since `hlist[domain]`  is not unavailable in it's scope as domain is what the function tries to return. I am currently looking into that.

quic-bkumar: I have added contains_of macro in one of my patches. Will get this merged soon. You can use this macro to get the structure pointer of the local handle available.

PR #30, opened 2024-07-31
Enable CI/CD pipeline with GitHub Actions to verify Build

quic-mtharu:
CI workflow to compile for Android and Linaro when a GitHub events "Pull request" or "Push" occurs on a GitHub-hosted virtual machines.


.github/workflows/compile_LE.yml
```diff
@@ -0,0 +1,58 @@
+name: CI - LE Compilation
```
ekanshibu: instead of LE, I think it's better to use ARM compilation with Linaro tools, same comment for the .yml file name

quic-ekangupt (approved):
LGTM

PR #38, opened 2024-08-07
adsprpc: PD tear down perf optimization

quic-rnallago:
Currently, after the last handle of a session is closed, the session is
also terminated. Instead, skip closing of the last handle and directly
terminate the remote session as this will reduce latency of session
cleanup.

### DIFF
--- a/src/fastrpc_apps_user.c
+++ b/src/fastrpc_apps_user.c
@@ -717,7 +717,7 @@ int get_domain_from_handle(remote_handle64 local, int *domain) {
 
   VERIFY(AEE_SUCCESS == (nErr = verify_local_handle(-1, local)));
   dom = (int)(hinfo->hlist - &hlist[0]);
-  VERIFYM((dom >= 0) && (dom < NUM_DOMAINS_EXTEND), AEE_EINVHANDLE,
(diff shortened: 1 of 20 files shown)


src/fastrpc_apps_user.c
```diff
@@ -1828,38 +1815,45 @@ int remote_handle64_close(remote_handle64 handle) {
...
+   * Terminate remote session if
+   *     1. there are no open non-domain handles AND
+   *     2. there are no open multi-domain handles, OR
+   *        only 1 multi-domain handle is open (for perf reason,
+   *        skip closing of it)
+   */
+  if (hlist[domain].domainsCount <= 1 && !hlist[domain].nondomainsCount)
+    start_deinit = true;
+  /*
+   * If session termination is not initiated and the remote handle is valid,
```
quic-bkumar: say a parallel high priority thread run remote_handle_open which increments this domainscount. remote_handle_close would try to close all these open handles. which is not an expected behavior.

quic-rnallago: The existing code approach in remote_handle64_close exhibits this issue, and I will create a new patch to address it.

PR #45, opened 2024-08-08
Add changes to support CDSP1 offloading

quic-lxu5:
FastRPC library supports 4 domains. There are some products where a new domain, CDSP1, is supported. Add changes to support CDSP1 domain.

### DIFF
--- a/inc/fastrpc_common.h
+++ b/inc/fastrpc_common.h
@@ -19,13 +19,13 @@
 
 /* Number of subsystem supported by fastRPC*/
 #ifndef NUM_DOMAINS
-#define NUM_DOMAINS 4
+#define NUM_DOMAINS 5
 #endif /*NUM_DOMAINS*/
 
 /* Number of sessions allowed per process */
 #ifndef NUM_SESSIONS
 #define NUM_SESSIONS 4
-#define DOMAIN_ID_MASK 3
+#define DOMAIN_ID_MASK 7
 #endif /*NUM_SESSIONS*/
 
 /* Default domain id, in case of non domains*/
--- a/src/fastrpc_apps_user.c
+++ b/src/fastrpc_apps_user.c
@@ -133,7 +133,7 @@ static void check_multilib_util(void);
 
 /* Array to store fastrpc library names. */
 static const char *fastrpc_library[NUM_DOMAINS] = {
-    "libadsprpc.so", "libmdsprpc.so", "libsdsprpc.so", "libcdsprpc.so"};
+    "libadsprpc.so", "libmdsprpc.so", "libsdsprpc.so", "libcdsprpc.so", "libcdsprpc.so"};
 
 /* Array to store env variable names. */
 static char *fastrpc_dsp_lib_refcnt[NUM_DOMAINS];
@@ -259,7 +259,7 @@ const char *ANDROID_DEBUG_VAR_NAME[] = {"fastrpc.process.attrs",
                                         "persist.fastrpc.process.attrs",
                                         "ro.build.type"};
 
-const char *SUBSYSTEM_NAME[] = {"adsp", "mdsp", "sdsp", "cdsp"};
+const char *SUBSYSTEM_NAME[] = {"adsp", "mdsp", "sdsp", "cdsp", "cdsp1"};
 
 /* Strings for trace event logging */
 #define INVOKE_BEGIN_TRACE_STR "fastrpc_msg: userspace_call: begin"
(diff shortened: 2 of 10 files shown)


src/fastrpc_apps_user.c
```diff
@@ -83,7 +83,7 @@
 #define DSP_DOM_LOCATION "/dsp/xdsp"
 #else
 #define DSP_MOUNT_LOCATION "/usr/lib/dsp/"
-#define DSP_DOM_LOCATION "/usr/lib/dsp/xdsp"
+#define DSP_DOM_LOCATION "/usr/lib/dsp/xdspn"
```
ekanshibu: I see this definition is already there in inc/apps_std_internal.h, can you please clean this up?

inc/fastrpc_common.h
```diff
@@ -19,13 +19,21 @@
 
 /* Number of subsystem supported by fastRPC*/
 #ifndef NUM_DOMAINS
+#ifdef MULTINSP_SUPPORT
```
ekanshibu: @quic-bkumar I think the featurization can be removed in github project, what do you suggest?

src/Makefile.am
```diff
@@ -91,6 +91,19 @@ libcdsp_default_listener_la_DEPENDENCIES = libcdsprpc.la
 libcdsp_default_listener_la_LDFLAGS = libcdsprpc.la -shared -module -avoid-version $(USE_LOG)
 libcdsp_default_listener_la_CFLAGS = $(CDSP_CFLAGS) -DUSE_SYSLOG
 
+CDSP1_CFLAGS = $(LIBDSPRPC_CFLAGS) -DDEFAULT_DOMAIN_ID=4
+
+lib_LTLIBRARIES += libcdsp1rpc.la
```
ekanshibu: can't you re-use libcdsprpc.so and cdsprpcd for CDSP1

src/fastrpc_apps_user.c
```diff
@@ -133,7 +133,7 @@ static void check_multilib_util(void);
 
 /* Array to store fastrpc library names. */
 static const char *fastrpc_library[NUM_DOMAINS] = {
-    "libadsprpc.so", "libmdsprpc.so", "libsdsprpc.so", "libcdsprpc.so"};
+    "libadsprpc.so", "libmdsprpc.so", "libsdsprpc.so", "libcdsprpc.so", "libcdsp1rpc.so"};
```
ekanshibu: why is libcdsp1rpc.so needed? Why not re-use libcdsprpc.so?

src/fastrpc_apps_user.c
```diff
@@ -2971,6 +2977,9 @@ int get_domain_from_name(const char *uri, uint32_t type) {
     } else if (!std_strncmp(uri, SDSP_DOMAIN_NAME,
                             std_strlen(SDSP_DOMAIN_NAME))) {
       domain = SDSP_DOMAIN_ID;
+    } else if (!std_strncmp(uri, CDSP1_DOMAIN_NAME,
+                            std_strlen(CDSP1_DOMAIN_NAME))) {
+      domain = CDSP1_DOMAIN_ID;
```
ekanshibu, line 2982: move this after CDSP_DOMAIN condition

src/fastrpc_apps_user.c
```diff
@@ -2987,6 +2996,8 @@ int get_domain_from_name(const char *uri, uint32_t type) {
       domain = MDSP_DOMAIN_ID;
     } else if (std_strstr(uri, SDSP_DOMAIN)) {
       domain = SDSP_DOMAIN_ID;
+    } else if (std_strstr(uri, CDSP1_DOMAIN)) {
+      domain = CDSP1_DOMAIN_ID;
```
ekanshibu, line 3000: move this after CDSP_DOMAIN condition

src/fastrpc_apps_user.c
quic-lxu5: > move this after CDSP_DOMAIN condition
Here we use std_strncmp, if we move CDSP before CDSP1, it will always match "cdsp" when we use "cdsp1" , as we only compare first four characters.

src/fastrpc_apps_user.c
quic-lxu5: same reason as above.
cdsp is sub-string of cdsp1, if we move cdsp above, it will always match cdsp even we need cdsp1.

src/Makefile.am
quic-bkumar: Agree, we should use libcdsprpc.so.

src/Makefile.am
```diff
@@ -91,6 +91,19 @@ libcdsp_default_listener_la_DEPENDENCIES = libcdsprpc.la
...
+libcdsp1rpc_la_SOURCES = $(LIBDSPRPC_SOURCES)
+libcdsp1rpc_la_LDFLAGS = -ldl -lm $(USE_LOG)
+libcdsp1rpc_la_CFLAGS = $(CDSP1_CFLAGS)
+
+lib_LTLIBRARIES += libcdsp1_default_listener.la
+libcdsp1_default_listener_la_SOURCES = $(LIBDEFAULT_LISTENER_SOURCES)
+libcdsp1_default_listener_la_DEPENDENCIES = libcdsp1rpc.la
+libcdsp1_default_listener_la_LDFLAGS = libcdsp1rpc.la -shared -module -avoid-version $(USE_LOG)
+libcdsp1_default_listener_la_CFLAGS = $(CDSP1_CFLAGS) -DUSE_SYSLOG
+
```
quic-bkumar: Same applies here. Better to use the cdsp default listener.

PR #52, opened 2024-08-11
Resolving the buffer accessed out of bound issue

quic-anankulk:
While packing shared buffer with names of all shared objects present in custom DSP_LIBRARY_PATH , their is a possibility of buffer overflow if the shared object names are exceeding the desired limit. The change makes sure the limit is not exceeded thus avoiding buffer overflow. Also, the buffer was allocated with 1KB memory which might fall short to accommodate all the needed shared object names so, increasing this size to 2KB.


quic-bkumar:
Anand, pls check the compilation failures.

quic-ekangupt:
@quic-anankulk can you resolve this compilation issue?

PR #53, opened 2024-08-12
Convert DSP error codes to user error codes

quic-bkumar:
Currently HLOS & DSP error logs are not aligned as these error codes get overwritten at DSP before returning back to HLOS. Move error codes conversion from DSP to HLOS in order to align error logs.


quic-bkumar:
Dropping this and bringing in another PR.

PR #55, opened 2024-08-14
adsprpc: Check fastrpc device existence before starting the poll

quic-rnallago:
In the existing code, there is a chance that the fastrpc device may already be present before the poll begins. This can lead to the fastrpc device event fails, resulting in a wait until the poll times out. Check the fastrpc device prior to initiating the poll.


quic-ekangupt:
Ramesh I see 5 commits with this PR, can you pls squash all to a single commit?

PR #56, opened 2024-08-14
adsprpc: Decrease the timeout for event polling and retry to poll

quic-rnallago:
Reduce timeout duration for event polling and retry the polling to avoid
a race condition between the device check and the polling process.


quic-ekangupt:
9 commits on this PR, pls squash all to a single commit

quic-rnallago:
This change not needed anymore

PR #60, opened 2024-09-26
Maintain correct domain handle count

quic-parihar:
In scenarios where static domain handle support is not available, wrong handle info is being passed to
fastrpc_update_module_list function. This leads to wrongful updation of domainsCount variable which is meant to keep count of dynamic domain handles. The corrupted domainsCount variable leads to premature exit of user process.


quic-ekangupt (approved):

quic-bkumar (commented):
removing perf, adspmsgd and current_process handles from the list needs similar changes.

quic-parihar:
Perf handle list change done

quic-ekangupt (approved):

quic-bkumar (approved):

PR #65, opened 2024-09-30
Handling fallback to non domain static handle calls

quic-parihar:
Currently in case of failure of static domain rpc calls we are falling back to non domain calls. The fallback to non domain calls should only happen when support for static domain based rpc calls is not present. Add proper check in error cases for static domain rpc calls.

### DIFF
--- a/src/fastrpc_apps_user.c
+++ b/src/fastrpc_apps_user.c
@@ -625,7 +625,7 @@ int fastrpc_set_remote_uthread_params(int domain) {
   if ((handle = get_remotectl1_handle(domain)) != INVALID_HANDLE) {
     nErr = remotectl1_set_param(handle, th_params->reqID, (uint32_t *)th_params,
                                 paramsLen);
-    if (nErr) {
+    if (nErr == DSP_AEE_EOFFSET + AEE_ENOSUCHMOD) {
       FARF(ALWAYS,
            "Warning 0x%x: %s: remotectl1 domains not supported for domain %d\n",
            nErr, __func__, domain);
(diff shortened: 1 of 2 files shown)


src/fastrpc_apps_user.c
```diff
@@ -625,7 +625,7 @@ int fastrpc_set_remote_uthread_params(int domain) {
   if ((handle = get_remotectl1_handle(domain)) != INVALID_HANDLE) {
     nErr = remotectl1_set_param(handle, th_params->reqID, (uint32_t *)th_params,
                                 paramsLen);
-    if (nErr) {
+    if (nErr == DSP_AEE_EOFFSET + AEE_ENOSUCHMOD) {
```
ekanshibu, line 628: if this call fails with ENOSUCHMOD, why in else case are we calling non-domain call? This means that if set_param fails due to some other reason, say BADPARM, we will again call non-domain call with bad parm. Ideally else case should just goto bail. Same comment for other places wherever applicable.

src/fastrpc_apps_user.c
```diff
@@ -1820,8 +1820,21 @@ int remote_handle_close_domain(int domain, remote_handle h) {
...
+               "%d\n",
+               nErr, __func__, domain);
+          fastrpc_update_module_list(DOMAIN_LIST_DEQUEUE, domain,
+                                     _const_remotectl1_handle, NULL, NULL);
+
+          // Set remotectlhandle to INVALID_HANDLE, so that all subsequent calls
+          // are non-domain calls
+          hlist[domain].remotectlhandle = INVALID_HANDLE;
+          nErr = remotectl_close(h, dlerrstr, err_str_len, &dlerr);
+        } else if (nErr)
```
ekanshibu, line 1836: here the handling looks correct

src/fastrpc_apps_user.c
quic-parihar: else condition is fallback for get_remotectl1_handle failure

quic-ekangupt (approved):

quic-bkumar (approved):

PR #85, opened 2024-10-21
Read all events and verify dev node after adding watcher fd

quic-abhishes:
Currently, the dev node is checked before adding the watcher FD. If the dev node is created between this check and the addition of the watcher FD, it will be missed by the poll. Therefore, check the dev node’s existence after adding the watcher FD. Additionally, read all events once the poll completes to ensure no events are missed.

### DIFF
--- a/src/adsp_default_listener.c
+++ b/src/adsp_default_listener.c
@@ -152,7 +152,7 @@ static int fastrpc_wait_for_secure_device(int domain)
 	while (1) {
 		int ret = 0;
 		char buffer[EVENT_BUF_LEN];
-		struct inotify_event* event;
+		struct inotify_event *event;
 
 		ret = poll(pfd, 1, POLL_TIMEOUT);
 		if(ret < 0){
--- a/src/adsp_default_listener.c
+++ b/src/adsp_default_listener.c
@@ -141,6 +141,10 @@ static int fastrpc_wait_for_secure_device(int domain)
 		VERIFY_EPRINTF("Error: inotify_add_watch failed, invalid fd errno = %s\n", strerror(errno));
 		return AEE_EINVALIDFD;
 	}
+
+	if (fastrpc_dev_exists(dev_name))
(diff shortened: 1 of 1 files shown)


src/adsp_default_listener.c
```diff
@@ -141,13 +141,18 @@ static int fastrpc_wait_for_secure_device(int domain)
 		VERIFY_EPRINTF("Error: inotify_add_watch failed, invalid fd errno = %s\n", strerror(errno));
 		return AEE_EINVALIDFD;
 	}
+
+	if (fastrpc_dev_exists(dev_name))
```
quic-ekangupt, line 145: there is 1 more similar check above, is that still needed?

src/adsp_default_listener.c
```diff
@@ -141,13 +141,18 @@ static int fastrpc_wait_for_secure_device(int domain)
...
+		goto bail;
+
 	memset(pfd, 0 , sizeof(pfd));
 	pfd[0].fd = inotify_fd;
 	pfd[0].events = POLLIN;
 
 	while (1) {
 		int ret = 0;
 		char buffer[EVENT_BUF_LEN];
+		struct inotify_event* event;
```
ekanshibu: struct inotify_event *event;

src/adsp_default_listener.c
ekanshibu, line 145: there is 1 more similar check above, is that still needed?

quic-abhishes: The first check helps to avoid the unnecessary creation and deletion of the inotify instance. The second check is essential to prevent race conditions between the creation of the device node and the addition of the watcher file descriptor. Therefore, both checks serve different purposes and are necessary.

quic-bkumar (approved):

quic-bkumar (approved):

quic-ekangupt (approved):

PR #86, opened 2024-10-21
Resolving the buffer accessed out of bound issue

anankulk:
While packing shared buffer with names of all shared objects present in custom DSP_LIBRARY_PATH , their is a possibility of buffer overflow if the shared object names are exceeding the desired limit.The change makes sure the limit is not exceeded thus avoiding buffer overflow. Also, the buffer was allocated with 1KB memory which might fall short to accomodate all the needed shared object names so, increasing this size to 2KB.

### DIFF
--- a/src/fastrpc_procbuf.c
+++ b/src/fastrpc_procbuf.c
@@ -22,6 +22,8 @@
 #define PROC_SHAREDBUF_SIZE (4*1024)
 #define ENV_PATH_LEN 256
 #define WORD_SIZE 4
+#define MAX_NON_PRELOAD_LIBS_LEN 2048
+#define FILE_EXT ".so"
 
 extern struct handle_list *hlist;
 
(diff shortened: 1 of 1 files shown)


src/fastrpc_procbuf.c
```diff
@@ -85,23 +89,28 @@ static int get_non_preload_lib_names (char** lib_names, size_t* buffer_size, int
...
-					size_t new_len = *buffer_size + std_strlen(file) + 1;
-					VERIFYC(NULL != (*lib_names = realloc (*lib_names, new_len)), AEE_ENOMEMORY);
-					std_strlcat(*lib_names, file, new_len);
-					std_strlcat(*lib_names, ";", new_len);
-					*buffer_size = new_len;
+				if (std_strstr(file, FILE_EXT) != NULL) {
+					if (concat_len + std_strlen(file) > MAX_NON_PRELOAD_LIBS_LEN) {
+						FARF(ALWAYS,"ERROR: Failed to pack library names in custom DSP_LIBRARY_PATH as required buffer size exceeds Max limit (%d).", MAX_NON_PRELOAD_LIBS_LEN);
+						nErr = AEE_EBUFFERTOOSMALL;
+						goto bail;
```
quic-mtharu, line 96: Please ensure to call closedir(dir) after opendir(path) to prevent a possible resource leak. This is crucial for proper resource management.

src/fastrpc_procbuf.c
```diff
@@ -69,13 +71,15 @@ int proc_sharedbuf_init(int dev, int domain) {
...
-	int nErr = AEE_SUCCESS, env_list_len = 0;
+	int nErr = AEE_SUCCESS, env_list_len = 0,  concat_len = 0, new_len = 0;
 	char* data_paths = NULL;
-	char *saveptr       = NULL;
+	char *saveptr = NULL;
+
+	VERIFYC(*lib_names != NULL, AEE_ENOMEMORY);
 	VERIFYC(NULL != (data_paths = calloc(1, sizeof(char) * ENV_PATH_LEN)), AEE_ENOMEMORY);
-	VERIFY(AEE_SUCCESS == (nErr = apps_std_getenv(DSP_LIBRARY_PATH, data_paths, ENV_PATH_LEN, &env_list_len)));
+	VERIFYC(AEE_SUCCESS == apps_std_getenv(DSP_LIBRARY_PATH, data_paths, ENV_PATH_LEN, &env_list_len), AEE_EGETENV);
```
quic-mtharu, line 80: instead of fixing the size of ENV_PATH_LEN to 256, Can we read the size of ENV_PATH_LEN by std_strlen(DSP_SEARCH_PATH)+1 and allocate the same amount of memory to data_paths

src/fastrpc_procbuf.c
```diff
@@ -22,6 +22,8 @@
 #define PROC_SHAREDBUF_SIZE (4*1024)
 #define ENV_PATH_LEN 256
```
quic-mtharu, line 23: Can we define all these macros in [apps_std_internal.h](https://github.com/quic/fastrpc/blob/bdb0bfee62f2210fa92a01b0e8d4132fac0fedea/inc/apps_std_internal.h#L20)
where our DSP_LIBRARY_PATH macro is defined.

quic-mtharu (changes requested):
Please ensure to call closedir(dir) after opendir(path) to prevent a possible resource leak. This is crucial for proper resource management.

Instead of fixing the size of ENV_PATH_LEN to 256, can we determine the size of ENV_PATH_LEN using strlen(DSP_SEARCH_PATH) + 1 and allocate the same amount of memory to data_paths?

Additionally, consider relocating all macros to the header file.

quic-ekangupt:
New PR created: https://github.com/quic/fastrpc/pull/113

PR #96, opened 2024-10-25
Handling fallback to non domain remote calls

quic-kartsana:
Currently fallback to non domain static handle calls is being done when we recieve module not supported error from DSP AEE_NOSUCHMOD. But, if module is not present the error code is being converted to AEE_ERPC. Replace fallback error code from AEE_NOSUCHMOD to AEE_ERPC.

### DIFF
--- a/src/fastrpc_apps_user.c
+++ b/src/fastrpc_apps_user.c
@@ -625,7 +625,7 @@ int fastrpc_set_remote_uthread_params(int domain) {
   if ((handle = get_remotectl1_handle(domain)) != INVALID_HANDLE) {
     nErr = remotectl1_set_param(handle, th_params->reqID, (uint32_t *)th_params,
                                 paramsLen);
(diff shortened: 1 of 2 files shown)


src/fastrpc_apps_user.c
```diff
@@ -625,7 +625,7 @@ int fastrpc_set_remote_uthread_params(int domain) {
   if ((handle = get_remotectl1_handle(domain)) != INVALID_HANDLE) {
     nErr = remotectl1_set_param(handle, th_params->reqID, (uint32_t *)th_params,
                                 paramsLen);
-    if (nErr == DSP_AEE_EOFFSET + AEE_ENOSUCHMOD) {
+    if (nErr == DSP_AEE_EOFFSET + AEE_ERPC) {
```
ekanshibu, line 628: AEE_ERPC can be returned for multiple reasons. Need to make it more accurate.

PR #111, opened 2024-11-15
Free memory allocated for storing pdname and runtime logs

quic-ansa:
There is a memory leak of 120 bytes every time handle open is requested for staticpd and function returns without freeing memory allocated for storing pdname in uri. Similarly 256 bytes of memory leaked with runtime logging buffer. Free this allocated memory before returning from function.

### DIFF
--- a/src/fastrpc_apps_user.c
+++ b/src/fastrpc_apps_user.c
@@ -1660,7 +1660,7 @@ int remote_handle_open_domain(int domain, const char *name, remote_handle *ph,
     } else if (!std_strncmp(pdName, "oispd", std_strlen("oispd"))) {
       hlist[domain].dsppd = OIS_STATICPD;
     }
-    return AEE_SUCCESS;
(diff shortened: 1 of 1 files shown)


quic-bkumar (approved):

src/fastrpc_apps_user.c
```diff
@@ -1660,7 +1660,7 @@ int remote_handle_open_domain(int domain, const char *name, remote_handle *ph,
     } else if (!std_strncmp(pdName, "oispd", std_strlen("oispd"))) {
       hlist[domain].dsppd = OIS_STATICPD;
     }
-    return AEE_SUCCESS;
+    goto free_buffer;
```
quic-mtharu, line 1663: Can't we use 'bail' only instead of new 'free_buffer' jump instructions ? 

Minimize 'goto' statements for better readability and maintainability.

PR #113, opened 2024-11-19
Resolving the buffer accessed out of bound issue.

quic-anankulk:
While packing shared buffer with names of all shared objects present in custom DSP_LIBRARY_PATH , their is a possibility of buffer overflow if the shared object names are exceeding the desired limit.The change makes sure the limit is not exceeded thus avoiding buffer overflow. Also, the buffer was allocated with 1KB memory which might fall short to accomodate all the needed shared object names so, increasing this size to 2KB.

### DIFF
--- a/src/fastrpc_procbuf.c
+++ b/src/fastrpc_procbuf.c
@@ -20,7 +20,6 @@
 
 /* size of buffer used to share the inital config params to dsp */
 #define PROC_SHAREDBUF_SIZE (4*1024)
-#define ENV_PATH_LEN 256
 #define WORD_SIZE 4
 
 extern struct handle_list *hlist;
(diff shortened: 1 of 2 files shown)


src/fastrpc_procbuf.c
```diff
@@ -85,23 +87,29 @@ static int get_non_preload_lib_names (char** lib_names, size_t* buffer_size, int
...
+						nErr = AEE_EBUFFERTOOSMALL;
+						closedir(dir);
+						goto bail;
+					}
+					std_strlcat(*lib_names, file, MAX_NON_PRELOAD_LIBS_LEN);
+					concat_len = std_strlcat(*lib_names, ";", MAX_NON_PRELOAD_LIBS_LEN);
 				}
 			}
 		}
 		closedir(dir);
```
quic-mtharu, line 102: Add a null check before calling closedir(dir) to ensure we are not attempting to close an invalid directory stream.

```
if (dir != NULL) {
    closedir(dir);
}
```

PR #115, opened 2024-11-22
Handle session close properly

quic-ekangupt:
Session close call is used by capability and context management functions. The handling should be in such a way that capability request should not close the device node if the PD is established. Handle session close correctly.

### DIFF
--- a/src/fastrpc_apps_user.c
+++ b/src/fastrpc_apps_user.c
@@ -362,12 +362,18 @@ int fastrpc_session_open(int domain, int *dev) {
   return AEE_ECONNREFUSED;
 }
 
-int fastrpc_session_close(int domain) {
-  int dev = hlist[domain].dev;
-
-  if (dev >= 0)
+void fastrpc_session_close(int domain, int dev) {
+  if (!hlist)
+    return;
+  if ((hlist[domain].dev == INVALID_DEVICE) &&
+      (dev != INVALID_DEVICE)) {
     close(dev);
-  return 0;
+  } else if ((hlist[domain].dev != INVALID_DEVICE) &&
+            (dev == INVALID_DEVICE)) {
+    close(hlist[domain].dev);
+    hlist[domain].dev = INVALID_DEVICE;
+  }
+  return;
 }
 
 int fastrpc_session_get(int domain) {
--- a/src/fastrpc_context.c
+++ b/src/fastrpc_context.c
@@ -134,7 +134,7 @@ static int fastrpc_context_deinit(fastrpc_context *ctx) {
 		if (!ctx->devs[i])
 			continue;
 
-		fastrpc_session_close(domain);
+		fastrpc_session_close(domain, INVALID_DEVICE);
 	}
 	free(ctx->devs);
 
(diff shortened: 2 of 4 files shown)


src/fastrpc_apps_user.c
```diff
@@ -362,11 +362,11 @@ int fastrpc_session_open(int domain, int *dev) {
   return AEE_ECONNREFUSED;
 }
 
-int fastrpc_session_close(int domain) {
-  int dev = hlist[domain].dev;
-
-  if (dev >= 0)
+int fastrpc_session_close(int domain, int dev) {
+  if (hlist[domain].dev == -1)
```
quic-mtharu: Can we check for the hlist before accessing it? to avoid null pointer dereferencing.

src/fastrpc_apps_user.c
```diff
@@ -362,11 +362,13 @@ int fastrpc_session_open(int domain, int *dev) {
   return AEE_ECONNREFUSED;
 }
 
-int fastrpc_session_close(int domain) {
-  int dev = hlist[domain].dev;
-
-  if (dev >= 0)
+int fastrpc_session_close(int domain, int dev) {
```
quic-mtharu: If the return value is not being checked in the calling function, should we consider changing the function's return type to void?

src/fastrpc_apps_user.c
```diff
@@ -362,11 +362,13 @@ int fastrpc_session_open(int domain, int *dev) {
...
+  if (!hlist)
+    return AEE_ENOTINITIALIZED;
+  if (hlist[domain].dev == -1)
     close(dev);
+  if (dev == -1 && hlist[domain].dev > 0)
+    close(hlist[domain].dev);
```
quic-mtharu, line 373: After closing the file descriptor, should we set hlist[domain].dev to -1 to avoid stale data in the hlist pointer?

src/fastrpc_apps_user.c
```diff
@@ -362,11 +362,13 @@ int fastrpc_session_open(int domain, int *dev) {
...
 
-int fastrpc_session_close(int domain) {
-  int dev = hlist[domain].dev;
-
-  if (dev >= 0)
+int fastrpc_session_close(int domain, int dev) {
+  if (!hlist)
+    return AEE_ENOTINITIALIZED;
+  if (hlist[domain].dev == -1)
     close(dev);
```
quic-mtharu, line 370: In case of the dev=-1 calling from [fastrpc_context.c], we are calling close() with -1.
Calling close(-1) will result in an error because -1 is not a valid file descriptor. The close() function will return -1 and set errno to EBADF (Bad file descriptor). It's important to ensure that the file descriptor is valid before attempting to close it

Change logic to if ((hlist[domain].dev == INVALID_DEVICE ) && (dev != INVALID_DEVICE))

INVALID_DEVICE is defined in fastrpc_common.h:40
#define INVALID_DEVICE (-1)

quic-mtharu (changes requested):

src/fastrpc_context.c
```diff
@@ -134,7 +134,7 @@ static int fastrpc_context_deinit(fastrpc_context *ctx) {
 		if (!ctx->devs[i])
 			continue;
 
-		fastrpc_session_close(domain);
+		fastrpc_session_close(domain, -1);
```
quic-mtharu: Pass this below macro, which we set as a part of fastrpc_common.h
INVALID_DEVICE

quic-mtharu (approved):

PR #134, opened 2025-01-27
Add changes for invokev2

quic-ekangupt:
Invoke V2 is going to support enhanced invocations like CRC, poll mode, etc. Complete documentation of invoke_v2 is added under Docs/ as part of this change.


lumag:
Please split the commit into logical chunks. It is impossible to review userspace changes if everything is thrown into a single bin.

PR #147, opened 2025-02-21
Resolve compilation failure

quic-ekangupt:
gettid() is causing implicit function declaration failure. Use syscall to get tid instead of gettid() call. Also add header to get definition of gettimeofday

### DIFF
--- a/inc/fastrpc_common.h
+++ b/inc/fastrpc_common.h
@@ -8,6 +8,7 @@
 #include <stdlib.h>
 #include <errno.h>
 #include <unistd.h>
+#include <sys/time.h>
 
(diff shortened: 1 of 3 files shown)


lumag:
- using `syscall()` means that you are bypassing possible glibc compatibility layers. Please don't do that. Instead read the man page for `gettid()`, it provides a way to solve 'implicit function declaration failure'.
- There is no connection between `gettid()` and `sys/time.h`. These changes should be coming as a separate changes.

quic-ekangupt:
This is getting addressed with a new PR: https://github.com/quic/fastrpc/pull/156

Closing this one.

PR #149, opened 2025-02-28
Add new function remotectl_open_v2

quic-vkatoch:
This commit introduces a new function, remotectl_open_v2, which includes additional arguments buf and bufLen. Previously, the dlopen_ex function executed with buf and bufLen set to zero on the DSP side. With this update, the virtual address and file length are now sourced from the userspace side.

### DIFF
--- a/src/mod_table.c
+++ b/src/mod_table.c
@@ -123,14 +123,14 @@ struct const_mod {
   char uri[1];
 };
 
-struct parsed_uri {
-  const char *file;
-  const char *sym;
-  const char *ver;
-  int filelen;
-  int symlen;
-  int verlen;
-};
+// struct parsed_uri {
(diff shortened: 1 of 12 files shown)


src/mod_table.c
```diff
@@ -123,14 +123,14 @@ struct const_mod {
...
 
-struct parsed_uri {
-  const char *file;
-  const char *sym;
-  const char *ver;
-  int filelen;
-  int symlen;
-  int verlen;
-};
+// struct parsed_uri {
```
kumarvbharath: Remove this..

quic-vkatoch: done

ekanshibu:
Moving PR to draft as this is a new method which requires DSP side implementation as well.

PR #154, opened 2025-03-18
refactor: platform-neutral API cleanup - rename Android-specific symbols and remove vendor path hardcoding

quic-vkatoch:
Remove Android-specific naming, dead code, and hardcoded paths from the fastrpc library to make it platform-neutral.
  - Rename listener_android.h/c to listener.h/c and update all symbol references throughout the codebase.
  - Replace fastrpc_android.h/c with fastrpc_platform.h/c exposing a neutral two-function API; move Android internals to file-static scope and add an LE stub in the same file.
  - Fix off-by-one OOB bounds check in fastrpc_get_property_int/string (> to >=) for the FASTRPC_BUILD_TYPE property key.
  - Remove hardcoded /vendor/dsp/ fallback paths from open_shell(); these are already covered by DEFAULT_DSP_SEARCH_PATHS at build time.
  - Fix header guard FASTRPC_ANDROID_USER_H to FASTRPC_APPS_USER_H.
  - Remove dead #if 0 blocks, stale comments, and Android-specific arrays that are no longer reachable after the platform layer refactor.

### DIFF
--- a/inc/AEEStdErr.h
+++ b/inc/AEEStdErr.h
@@ -84,7 +84,7 @@
 #define AEE_EINVALIDITEM              (AEE_EOFFSET + 0x02A)  ///< Current item is invalid, it can be a switch case or a pointer to memory
 #define AEE_ENOTALLOWED               (AEE_EOFFSET + 0x02B)  ///< Not allowed to perform the
                                                                     ///< operation
-#define AEE_EINVHANDLE                (AEE_EOFFSET + 0x02C)  ///< Invalid handle - adding here as its defined in vendor AEEStdErr.h - needed to check valid handle in stub.c
+#define AEE_EINVHANDLE                (AEE_EOFFSET + 0x02C)  ///< Invalid handle - needed to check valid handle in stub.c
 #define AEE_EOUTOFHANDLES             (AEE_EOFFSET + 0x02D)  ///< Out of handles (Handle list is already full)
 //Hole here
 #define AEE_ENOMORE                   (AEE_EOFFSET + 0x02F)  ///< No more items available --
--- a/inc/apps_std_internal.h
+++ b/inc/apps_std_internal.h
@@ -22,13 +22,6 @@
 #define MAX_NON_PRELOAD_LIBS_LEN 2048
 #define FILE_EXT ".so"
 
-#ifndef VENDOR_DSP_LOCATION
-#define VENDOR_DSP_LOCATION "/vendor/dsp/"
-#endif
-#ifndef VENDOR_DOM_LOCATION
-#define VENDOR_DOM_LOCATION "/vendor/dsp/xdsp/"
-#endif
-
 // Search path used by fastRPC for acdb path
 #ifndef ADSP_AVS_CFG_PATH
 #define ADSP_AVS_CFG_PATH ";/etc/acdbdata/;"
--- a/inc/apps_std_internal.h
+++ b/inc/apps_std_internal.h
@@ -21,12 +21,6 @@
 #define MAX_NON_PRELOAD_LIBS_LEN 2048
 #define FILE_EXT ".so"
 
-#ifndef VENDOR_DSP_LOCATION
-#define VENDOR_DSP_LOCATION "/vendor/dsp/"
-#endif
-#ifndef VENDOR_DOM_LOCATION
-#define VENDOR_DOM_LOCATION "/vendor/dsp/xdsp/"
-#endif
 
 int fopen_from_dirlist(const char *dirList, const char *delim, 
     const char *mode, const char *name, apps_std_FILE *psout);
--- a/inc/fastrpc_android.h
+++ b/inc/fastrpc_android.h
@@ -1,0 +1,17 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+
+#ifndef FASTRPC_ANDROID_H
+#define FASTRPC_ANDROID_H
+
+#include "fastrpc_common.h"
+
+/* Platform property system functions */
+int property_get_int32(const char *name, int value);
+int property_get(const char *name, int *def, int *value);
+
+/* Platform-specific property getter functions */
+int platform_get_property_int(fastrpc_properties UserPropKey, int defValue);
+int platform_get_property_string(fastrpc_properties UserPropKey, char *value, char *defValue);
+
+#endif /*FASTRPC_ANDROID_H*/
--- a/inc/fastrpc_apps_user.h
+++ b/inc/fastrpc_apps_user.h
@@ -1,8 +1,8 @@
 // Copyright (c) 2024, Qualcomm Innovation Center, Inc. All rights reserved.
 // SPDX-License-Identifier: BSD-3-Clause
 
-#ifndef FASTRPC_ANDROID_USER_H
-#define FASTRPC_ANDROID_USER_H
+#ifndef FASTRPC_APPS_USER_H
+#define FASTRPC_APPS_USER_H
 
 #include <assert.h>
 #include <fcntl.h>
@@ -134,4 +134,4 @@ int get_unsigned_pd_attribute(uint32_t domain, int *unsigned_module);
  */
 int is_userspace_allocation_supported();
 
-#endif //FASTRPC_ANDROID_USER_H
+#endif //FASTRPC_APPS_USER_H
--- a/inc/fastrpc_trace.h
+++ b/inc/fastrpc_trace.h
@@ -4,9 +4,6 @@
 #ifndef FASTRPC_TRACE_H
 #define FASTRPC_TRACE_H
 
-#if ((defined _ANDROID) || (defined ANDROID)) || (defined DISABLE_ATRACE) && !defined(LE_ENABLE)
-//TODO: Bharath #include "cutils/trace.h" //for systrace support
-#endif
 #include "HAP_farf.h"
 
 //for systrace support
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -54,7 +54,7 @@ LIBDSPRPC_SOURCES = \
 		fastrpc_context.c
 
 if ANDROID_CC
-# Do nothing (or Android-specific sources)
+LIBDSPRPC_SOURCES += fastrpc_android.c
 else
 LIBDSPRPC_SOURCES += fastrpc_config_parser.c
 endif
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -24,7 +24,7 @@ LIBDSPRPC_SOURCES = \
 		dspsignal.c \
 		dspqueue/dspqueue_cpu.c \
 		dspqueue/dspqueue_rpc_stub.c \
-		listener_android.c \
+		listener.c \
 		apps_std_imp.c \
 		apps_mem_imp.c \
 		apps_mem_skel.c \
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -52,11 +52,10 @@ LIBDSPRPC_SOURCES = \
 		adsp_perf_stub.c \
 		adsp_perf1_stub.c \
 		mod_table.c \
-		fastrpc_context.c
+		fastrpc_context.c \
+		fastrpc_platform.c
 
-if ANDROID_CC
-# Do nothing (or Android-specific sources)
-else
+if !ANDROID_CC
 LIBDSPRPC_SOURCES += fastrpc_config_parser.c
 endif
 
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -25,7 +25,7 @@ LIBDSPRPC_SOURCES = \
 		dspsignal.c \
 		dspqueue/dspqueue_cpu.c \
 		dspqueue/dspqueue_rpc_stub.c \
-		listener_android.c \
+		listener.c \
 		apps_std_imp.c \
 		apps_mem_imp.c \
 		apps_mem_skel.c \
--- a/src/fastrpc_android.c
+++ b/src/fastrpc_android.c
@@ -1,0 +1,58 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+
+#include "fastrpc_android.h"
+#include "HAP_farf.h"
+#include <string.h>
+
+#define PROPERTY_VALUE_MAX 92
+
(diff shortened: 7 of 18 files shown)


src/fastrpc_apps_user.c
```diff
@@ -3548,12 +3483,6 @@ static int open_shell(int domain_id, apps_std_FILE *fh, int unsigned_shell) {
     if (domain == SDSP_DOMAIN_ID && fh != NULL) {
       nErr = AEE_SUCCESS;
       *fh = -1;
-    } else {
-      FARF(ERROR,
```
ekanshibu, line 3441: should not completely remove the error log

src/adspmsgd.c
```diff
@@ -1,176 +1,176 @@
...
-}
+// Copyright (c) 2024, Qualcomm Innovation Center, Inc. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
+
+#ifndef VERIFY_PRINT_ERROR
+#define VERIFY_PRINT_ERROR
+#endif
+
+#include <string.h>
+#include <errno.h>
```
ekanshibu: can you highlight what is changed in this file?

quic-vkatoch: Nothing has changed in this file.

src/fastrpc_apps_user.c
quic-vkatoch: The error log contains vendor-specific location macros.

src/fastrpc_apps_user.c
```diff
@@ -241,16 +230,6 @@ const char *ENV_DEBUG_VAR_NAME[] = {"FASTRPC_PROCESS_ATTRS",
...
-                                         "vendor.fastrpc.debug.trace",
-                                         "vendor.fastrpc.debug.testsig",
-                                         "vendor.fastrpc.perf.kernel",
-                                         "vendor.fastrpc.perf.adsp",
-                                         "vendor.fastrpc.perf.freq",
-                                         "vendor.fastrpc.debug.systrace",
-                                         "vendor.fastrpc.debug.pddump",
-                                         "persist.vendor.fastrpc.process.attrs",
-                                         "ro.build.type"};
 const char *ANDROID_DEBUG_VAR_NAME[] = {"fastrpc.process.attrs",
```
lumag: This is also Android-specific var and was missed

quic-vkatoch: Thanks for pointing that out. I am working on the cleanup and will include it in the next patch.

src/fastrpc_apps_user.c
```diff
@@ -241,26 +230,6 @@ const char *ENV_DEBUG_VAR_NAME[] = {"FASTRPC_PROCESS_ATTRS",
                                     "FASTRPC_DEBUG_PDDUMP",
                                     "FASTRPC_PROCESS_ATTRS_PERSISTENT",
                                     "ro.debuggable"};
```
lumag: `ro.debuggable` is also an Android definition.

src/fastrpc_apps_user.c
lumag: So, the error log needs to be fixed rather than being completely removed.

quic-vkatoch: Okay, I'll only be logging the paths that have been actually checked.

src/fastrpc_apps_user.c
quic-vkatoch: I'll check and remove it.

quic-ekangupt:
Please check if you can break the change into multiple commits instead of a single commit.

src/fastrpc_log.c
```diff
@@ -288,17 +287,7 @@ void fastrpc_log_init() {
...
-    if (!strncmp(build_type, "eng", PROPERTY_VALUE_MAX) ||
-        !strncmp(build_type, "userdebug", PROPERTY_VALUE_MAX))
-      debug_build_type = true;
-#else
-    if (atoi(build_type))
-      debug_build_type = true;
-#endif
-  }
-  if (persist_buf.buf == NULL && debug_build_type) {
+  if (persist_buf.buf == NULL) {
```
lumag: These are separate changes. Why are you squashing them into a single commit?

quic-vkatoch: I have separated the changes into multiple commits and it's no longer a single commit. I'm checking if it could be separated any further.

quic-ekangupt:
@quic-vkatoch, can you please rebase this PR?

src/fastrpc_log.c
```diff
@@ -1,15 +1,13 @@
 // Copyright (c) 2024, Qualcomm Innovation Center, Inc. All rights reserved.
 // SPDX-License-Identifier: BSD-3-Clause
 
-#define _GNU_SOURCE
```
ekanshibu, line 4: why removing this?

inc/AEEStdErr.h
```diff
@@ -84,7 +84,7 @@
 #define AEE_EINVALIDITEM              (AEE_EOFFSET + 0x02A)  ///< Current item is invalid, it can be a switch case or a pointer to memory
 #define AEE_ENOTALLOWED               (AEE_EOFFSET + 0x02B)  ///< Not allowed to perform the
```
ekanshibu, line 85: fix all commit messages

ekanshibu (changes requested):

src/fastrpc_log.c
quic-vkatoch: Updated.

inc/AEEStdErr.h
quic-vkatoch: Fixed.

ekanshibu (dismissed):

inc/apps_std_internal.h
Chennak-quic, line 1: If we are removing the support for /vendor/dsp, where is shell expected to be pushed for android targets ? if it is /usr/lib/dsp/ like other LE targets,  is /usr partition available on android targets ? 

how is this change tested on android targets ?

inc/apps_std_internal.h
```diff
@@ -39,16 +39,9 @@
 #endif
 #endif /* ENABLE_UPSTREAM_DRIVER_INTERFACE */
 
-#ifndef VENDOR_DSP_LOCATION
```
lumag, line 24: How are these two changes related to each other?

inc/fastrpc_trace.h
```diff
@@ -4,9 +4,6 @@
 #ifndef FASTRPC_TRACE_H
 #define FASTRPC_TRACE_H
 
-#if ((defined _ANDROID) || (defined ANDROID)) || (defined DISABLE_ATRACE) && !defined(LE_ENABLE)
```
lumag, line 7: Again, totally unrelated changes

src/pl_list.c
```diff
@@ -4,7 +4,7 @@
 #include "platform_libs.h"
 
 PL_DEP(gpls)
-PL_DEP(listener_android)
+PL_DEP(listener)
```
lumag, line 7: Sould this be a part of the previous commit?

src/Makefile.am
```diff
@@ -19,7 +19,7 @@ LIBDSPRPC_SOURCES = \
 		dspsignal.c \
 		dspqueue/dspqueue_cpu.c \
 		dspqueue/dspqueue_rpc_stub.c \
-		listener_android.c \
+		listener.c \
```
lumag, line 28: Sould this be a part of the previous commit?

src/fastrpc_apps_user.c
```diff
@@ -64,7 +64,7 @@
 #include "fastrpc_perf.h"
 #include "fastrpc_pm.h"
 #include "fastrpc_procbuf.h"
-#include "listener_android.h"
+#include "listener.h"
```
lumag, line 65: Ugh. Make sure that each commit works on its own. The code was broken up to now.

src/fastrpc_apps_user.c
```diff
@@ -4102,7 +4002,7 @@ static void fastrpc_apps_user_deinit(void) {
       pthread_mutex_destroy(&hlist[i].init);
       pthread_mutex_destroy(&hlist[i].async_init_deinit_mut);
     }
-    listener_android_deinit();
+    listener_deinit();
```
lumag, line 3869: Yuck

src/fastrpc_log.c
```diff
@@ -162,44 +162,59 @@ void HAP_debug_runtime(int level, const char *file, int line,
                        const char *format, ...) {
   int len = 0;
   va_list argp;
-  char *buf = NULL, *log = NULL;
```
lumag, line 165: Again, feels like several unrelated changed squashed toggether.

lumag (changes requested):

inc/apps_std_internal.h
quic-vkatoch: These changes are unrelated; I will update the PR.

inc/fastrpc_trace.h
quic-vkatoch: These changes are unrelated; I will update the PR.

src/Makefile.am
quic-vkatoch: You're right, I'll do that.

src/fastrpc_apps_user.c
quic-vkatoch: Sure, I'll do that.

src/fastrpc_apps_user.c
quic-vkatoch: I will update the PR.

src/pl_list.c
quic-vkatoch: You're right, I'll do that.

src/fastrpc_log.c
quic-vkatoch: When I removed debug_build_type (ro.debuggable) under fastrpc_log_init, it caused issues due to incorrectness in the HAP_debug_runtime function, so I had to update it. I'll remove this change from the current PR and submit a separate PR to address the fix.

inc/apps_std_internal.h
quic-vkatoch: I have to check where the shell should be placed if /vendor/dsp support is removed, or if we actually need to remove it in the first place.

Regarding testing: earlier, the daemons were already running (opened the shell) and the tests passed. After killing and updating with new daemons, the tests are now failing. I will check this further and update.

lumag: The shell is loaded from the same folder as all other DSP libraries. What's the question?

lumag: > If we are removing the support for /vendor/dsp, where is shell expected to be pushed for android targets ? if it is /usr/lib/dsp/ like other LE targets, is /usr partition available on android targets ?

All DSP binaries are located in the subdirs of `/usr/share/qcom`, one subdir per device.

lumag: On Android systems you can reuse the same approach, changing /usr/share/qcom/ to /vendor or its subdir. All config files should be relative to that path,

quic-bkumar (dismissed):

quic-vkatoch:
@lumag, could you please review this PR?

src/fastrpc_android.c
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) 2024, Qualcomm Innovation Center, Inc. All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include "fastrpc_android.h"
```
ekanshibu: adding empty file? also, fix the copyright, refer to https://github.com/qualcomm/fastrpc/blob/development/src/fastrpc_config_parser.c#L1

inc/fastrpc_android.h
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2024, Qualcomm Innovation Center, Inc. All rights reserved.
```
ekanshibu: fix copyright, ref: https://github.com/qualcomm/fastrpc/blob/development/src/fastrpc_config_parser.c#L1

ekanshibu (changes requested):

src/fastrpc_android.c
ekanshibu: By empty file, I mean the empty file is added as part of 1st commit, is that expected?

inc/fastrpc_apps_user.h
```diff
@@ -1,8 +1,8 @@
 // Copyright (c) 2024, Qualcomm Innovation Center, Inc. All rights reserved.
 // SPDX-License-Identifier: BSD-3-Clause
 
-#ifndef FASTRPC_ANDROID_USER_H
-#define FASTRPC_ANDROID_USER_H
+#ifndef FASTRPC_USER_H
```
lumag: Then it should be FASTRPC_APPS_USER_H

inc/fastrpc_android.h
```diff
@@ -4,6 +4,19 @@
...
+#include "fastrpc_common.h"
+
 #define DEFAULT_DSP_SEARCH_PATHS ";/vendor/lib/rfsa/adsp;/vendor/dsp/;"
 
+/* Android property system functions */
+int property_get_int32(const char *name, int value);
+int property_get(const char *name, int *def, int *value);
+
+/* Android-specific property getter functions */
+int android_get_property_int(fastrpc_properties UserPropKey, int defValue);
```
lumag: Drop `android_foo()` functions from the public header. Define platform-neutral API, which can be implemented for Android, LE, or some other OS.

src/fastrpc_android.c
```diff
@@ -1,4 +1,127 @@
...
+int property_get_int32(const char *name, int value) {
+    return 0;
+}
+
+int property_get(const char *name, int *def, int *value) {
+    return 0;
+}
+
+int android_get_property_int(fastrpc_properties UserPropKey, int defValue) {
+#if !defined(LE_ENABLE) // Android platform
```
lumag: !LE_ENABLE can be dropped

lumag (changes requested):

src/fastrpc_android.c
quic-vkatoch: Done. Moved to a later commit.

inc/fastrpc_android.h
quic-vkatoch: Ack, updated.

inc/fastrpc_android.h
quic-bkumar, line 1: Are we okay to name the file with android in it?

Chennak-quic (dismissed):

inc/fastrpc_android.h
quic-vkatoch: Updated, to make it platform-neutral

PR #159, opened 2025-04-15
Add changes to support GPDSPs offloading

quic-lxu5:
FastRPC library supports 5 domains. There are some products where new domains, GPDSP0 and GPDSP1, are supported. Add changes to support GPDSP domains.


lumag:
Hmm, why?

PR #163, opened 2025-04-21
Install fastrpc headers to installdir

quic-ekangupt:
Few of the fastrpc header files are used by developers for their projects. Copy these headers from source directory to target directory to ensure that they are available for developers to include in their own projects.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -129,5 +129,20 @@ sdsprpcd_CFLAGS = -I$(top_srcdir)/inc -DDEFAULT_DOMAIN_ID=2 -DUSE_SYSLOG -DNO_HA
 sdsprpcd_LDADD =  -ldl $(USE_LOG)
 
 
-otherincludedir = $(includedir)/fastrpc
-otherinclude_HEADERS = $(top_srcdir)/inc/*.h
+# Export fastrpc headers
+fastrpc_includedir       = $(includedir)
+fastrpc_include_HEADERS  = $(top_srcdir)/inc/AEEatomic.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/AEEQList.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/AEEstd.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/AEEStdDef.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/AEEStdErr.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/AEEVaList.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/remote.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/remote64.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/rpcmem.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/verify.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/HAP_farf.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/HAP_pls.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/HAP_debug.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/dspqueue.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/adsp_default_listener1.h


src/Makefile.am
```diff
@@ -129,5 +129,20 @@ sdsprpcd_CFLAGS = -I$(top_srcdir)/inc -DDEFAULT_DOMAIN_ID=2 -DUSE_SYSLOG -DNO_HA
...
+fastrpc_include_HEADERS += $(top_srcdir)/inc/AEEVaList.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/remote.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/remote64.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/rpcmem.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/verify.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/HAP_farf.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/HAP_pls.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/HAP_debug.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/dspqueue.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/adsp_default_listener1.h
```
lumag: Which of the headers actually define an API that is useable by the users of fastrpc?

ekanshibu: All these headers are maintained as shippable components and are utilized by various clients. Specifically, for QLI supporting platforms, some headers are not used. However, generally, all headers are provided for users of fastrpc.

src/Makefile.am
```diff
@@ -129,5 +129,20 @@ sdsprpcd_CFLAGS = -I$(top_srcdir)/inc -DDEFAULT_DOMAIN_ID=2 -DUSE_SYSLOG -DNO_HA
 sdsprpcd_LDADD =  -ldl $(USE_LOG)
 
 
-otherincludedir = $(includedir)/fastrpc
-otherinclude_HEADERS = $(top_srcdir)/inc/*.h
+# Export fastrpc headers
+fastrpc_includedir       = $(includedir)
+fastrpc_include_HEADERS  = $(top_srcdir)/inc/AEEatomic.h
```
lumag: No. We don't need another atomic wrapper, users should use -latomic instead. Please don't export this one

src/Makefile.am
```diff
@@ -129,5 +129,20 @@ sdsprpcd_CFLAGS = -I$(top_srcdir)/inc -DDEFAULT_DOMAIN_ID=2 -DUSE_SYSLOG -DNO_HA
...
+fastrpc_include_HEADERS += $(top_srcdir)/inc/AEEQList.h
```
lumag: Yet another list header. It is not a DSP interface, so please drop it from the exported headers. If necessary, people can use [<sys/queue.h>](https://man7.org/linux/man-pages/man3/queue.3.html)

src/Makefile.am
```diff
@@ -129,5 +129,20 @@ sdsprpcd_CFLAGS = -I$(top_srcdir)/inc -DDEFAULT_DOMAIN_ID=2 -DUSE_SYSLOG -DNO_HA
...
+fastrpc_include_HEADERS += $(top_srcdir)/inc/AEEstd.h
```
lumag: A random collection of libc wrappers and some extra functions. Drop it and use libc directly. Don't encourage our users to use unnecessary libc wrappers.

src/Makefile.am
```diff
@@ -129,5 +129,20 @@ sdsprpcd_CFLAGS = -I$(top_srcdir)/inc -DDEFAULT_DOMAIN_ID=2 -DUSE_SYSLOG -DNO_HA
...
+fastrpc_include_HEADERS += $(top_srcdir)/inc/AEEStdDef.h
```
lumag: This is for those that don't know about stddef.h / stdint.h. Please drop it completely.

src/Makefile.am
```diff
@@ -129,5 +129,20 @@ sdsprpcd_CFLAGS = -I$(top_srcdir)/inc -DDEFAULT_DOMAIN_ID=2 -DUSE_SYSLOG -DNO_HA
...
+fastrpc_include_HEADERS += $(top_srcdir)/inc/AEEStdErr.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/AEEVaList.h
```
lumag: Use stdarg.h instead, drop the header.

src/Makefile.am
```diff
@@ -129,5 +129,20 @@ sdsprpcd_CFLAGS = -I$(top_srcdir)/inc -DDEFAULT_DOMAIN_ID=2 -DUSE_SYSLOG -DNO_HA
...
+fastrpc_include_HEADERS += $(top_srcdir)/inc/remote.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/remote64.h
```
lumag: "remote64.h will be deleted in future". Please stop exporting it.

src/Makefile.am
```diff
@@ -129,5 +129,20 @@ sdsprpcd_CFLAGS = -I$(top_srcdir)/inc -DDEFAULT_DOMAIN_ID=2 -DUSE_SYSLOG -DNO_HA
...
+fastrpc_include_HEADERS += $(top_srcdir)/inc/rpcmem.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/verify.h
```
lumag: This is an _internal_ header, don't export it to outside users.

src/Makefile.am
```diff
@@ -129,5 +129,20 @@ sdsprpcd_CFLAGS = -I$(top_srcdir)/inc -DDEFAULT_DOMAIN_ID=2 -DUSE_SYSLOG -DNO_HA
...
+fastrpc_include_HEADERS += $(top_srcdir)/inc/HAP_farf.h
```
lumag: This one also looks like an internal header, rather than something that should be used by library users.

src/Makefile.am
```diff
@@ -129,5 +129,20 @@ sdsprpcd_CFLAGS = -I$(top_srcdir)/inc -DDEFAULT_DOMAIN_ID=2 -DUSE_SYSLOG -DNO_HA
...
+fastrpc_include_HEADERS += $(top_srcdir)/inc/HAP_pls.h
```
lumag: >  * Warning, this API should only be called from within a thread started by FastRPC, and not from any user created threads via the qurt apis.

So, this is not the API for apps.

src/Makefile.am
```diff
@@ -129,5 +129,20 @@ sdsprpcd_CFLAGS = -I$(top_srcdir)/inc -DDEFAULT_DOMAIN_ID=2 -DUSE_SYSLOG -DNO_HA
...
+fastrpc_include_HEADERS += $(top_srcdir)/inc/HAP_debug.h
```
lumag: Is it a debug for adsprpcd or for user apps?

src/Makefile.am
lumag: `adsp_default_listener_register` is not documented at all. What is it?

lumag (changes requested):

src/Makefile.am
lumag: I quickly went through the headers. Some of them are dsprpcd-internal headers. Some are bad wrappers which should have been dropped long ago. So, no, please don't randomly export those headers.

ekanshibu: All these headers are already provided in **dspservices-headers_15.0.qcom_aarch64.tar.gz** prebuilt (https://docs.qualcomm.com/bundle/publicresource/topics/RNO-250403001134/ReleaseNote.html?vproduct=1601111740013072&version=1.4)

I'm checking the impact on clients if I restrict exporting any headers.

src/Makefile.am
ekanshibu: this header is used by sensors and audio daemons.

src/Makefile.am
ekanshibu: I see this is included by multiple clients like sensors, CV etc.

src/Makefile.am
ekanshibu: I'll add documentation for the header as a separate commit.

src/Makefile.am
lumag: Why? It's definitely an internal header, depending on the implementation details (like the use of liblog, FARF, syslog or just printf). If any app wants to use it, it should vendor in this header: it doesn't provide an API, it is not directly related to the fastrpc, etc. So, as the fastrpc wasn't installing any headers beforehand, it should be excluded.

quic-ekangupt:
accidently closed this PR

quic-ekangupt:
I'm planning to install only the following headers:
- rpcmem.h(To export rpcmem_alloc which can be used by clients)
- remote.h(remote interfaces and APIs)
- AEEStdErr.h(Error codes used in fastrpc framework)

src/Makefile.am
```diff
@@ -129,5 +129,8 @@ sdsprpcd_CFLAGS = -I$(top_srcdir)/inc -DDEFAULT_DOMAIN_ID=2 -DUSE_SYSLOG -DNO_HA
 sdsprpcd_LDADD =  -ldl $(USE_LOG)
 
 
-otherincludedir = $(includedir)/fastrpc
-otherinclude_HEADERS = $(top_srcdir)/inc/*.h
+# Export fastrpc headers
+fastrpc_includedir       = $(includedir)
+fastrpc_include_HEADERS += $(top_srcdir)/inc/AEEStdErr.h
```
lumag: Just `include_HEADERS += $(top_srcdir)/inc/AEEStdErr.h`, etc.

lumag: Wait... No. The previous version was better. Don't stuff things into `${includedir}`, use a subdir like a previous implementation did.

ekanshibu: I think I have messed up the PR, fixing this

lumag (approved):

quic-mtharu (approved):

Chennak-quic (approved):

PR #168, opened 2025-04-25
Create fastrpc-lava-tests.yml

quic-mtharu:
Testing purpose


mwasilew (changes requested):
I'm missing the download URL. Maybe I'm missing how the build is supposed to work.

PR #169, opened 2025-04-25
test: install testing libraries

lumag:
In order to be able to execute a test on the target, install ARM64 libraries and DSP files.


quic-ekangupt (approved):
tested on linux

quic-bkumar (approved):

PR #170, opened 2025-04-30
fix: Link test libs to corrected version libcdsprpc.so.1

quic-vkatoch:
Ensured test libraries are linked to the correct versioned library libcdsprpc.so.1


lumag (approved):
Thanks!

quic-ekangupt (approved):

quic-bkumar (approved):

PR #182, opened 2025-06-03
Add GitHub Actions workflow for FastRPC build and QEMU test with hello message

quic-mtharu:
This commit introduces a GitHub Actions workflow that automates the following:

- Clones and builds the FastRPC library
- Compiles a simple test binary that prints "hello"
- Packages the binary into an initramfs with a minimal init script
- Builds a Linux kernel and boots it using QEMU
- Captures QEMU boot logs and verifies the presence of the "hello" message

This setup ensures that the FastRPC environment is correctly built and that basic test execution is validated during the virtual boot process.


quic-mtharu:
Direct push is restricted on the branch, so will create a separate PR from the forked project.

PR #194, opened 2025-07-14
test: Update environment handling and README for fastrpc_test

quic-vkatoch:
- Update fastrpc_test.c to append testlibdir and testdspdir to LD_LIBRARY_PATH and DSP_LIBRARY_PATH instead of overwriting them.
- Modify Makefile.am to pass -Dtestlibdir and -Dtestdspdir for correct environment variable setup.
- Revise README.md to use make install for installation and simplify run instructions, removing the need to change directories.

### DIFF
--- a/test/Makefile.am
+++ b/test/Makefile.am
@@ -5,7 +5,9 @@ bin_PROGRAMS = fastrpc_test
 fastrpc_test_SOURCES = fastrpc_test.c
 
 # Define the compiler flags for the test program
-fastrpc_test_CFLAGS = -I$(top_srcdir)/inc -DUSE_SYSLOG
+fastrpc_test_CFLAGS = -I$(top_srcdir)/inc -DUSE_SYSLOG \
+                      -Dtestlibdir=\"$(testlibdir)\" \
+                      -Dtestdspdir=\"$(testdspdir)\"
 
 if ANDROID_CC
 USE_LOG = -llog
--- a/test/README.md
+++ b/test/README.md
@@ -12,28 +12,20 @@ After building the application, the following files and directories need to be p
 4. **v68 Directory**: Contains skeletons for the v68 architecture version.
 5. **v75 Directory**: Contains skeletons for the v75 architecture version.
 
-Copy the following files and directories to `/usr/bin` on the target device:
+### Installing the Test
 
-- `/path/to/your/fastrpc/test/.libs/fastrpc_test`
-- `/path/to/your/fastrpc/test/android` (if using Android)
-- `/path/to/your/fastrpc/test/linux` (if using Linux)
-- `/path/to/your/fastrpc/test/v68`
-- `/path/to/your/fastrpc/test/v75`
+Use `make install` to install the libraries and the executable to the appropriate directories:
 
-**Note**: Push the `android` directory if you are using the Android platform, and the `linux` directory if you are using the Linux platform.
+```bash
+make install
+```
 
 ## Running the Test
 
-To run the test application, follow these steps:
-
-1. Navigate to the `/usr/bin` directory on the device.
-2. Execute the `fastrpc_test` binary with the appropriate options.
-
-Example command:
+To run the test application, use the following command:
 
 ```bash
-cd /usr/bin
-./fastrpc_test -d 3 -U 1 -t linux -a v68
+fastrpc_test -d 3 -U 1 -t linux -a v68
 ```
 
 ### Options
--- a/test/fastrpc_test.c
+++ b/test/fastrpc_test.c
@@ -39,9 +39,10 @@ int main(int argc, char *argv[]) {
     bool is_unsignedpd_enabled = true;  // Default to unsigned PD
     const char *target = "linux";  // Default target platform
     const char *arch_version = "v68";  // Default architecture version
-    char abs_lib_path[PATH_MAX];
     char ld_lib_path[PATH_MAX];
     char dsp_lib_path[PATH_MAX];
+    char *current_ld_lib_path;
+    char *current_dsp_lib_path;
     DIR *dir;
     struct dirent *entry;
     char full_lib_path[PATH_MAX];
@@ -60,6 +61,11 @@ int main(int argc, char *argv[]) {
                 break;
             case 't':
                 target = optarg;
+                if (strcmp(target, "linux") != 0 && strcmp(target, "android") != 0) {
+                    printf("\nERROR: Invalid target platform (-t). Must be linux or android.\n");
+                    print_usage();
+                    return -1;
+                }
                 break;
             case 'a':
                 arch_version = optarg;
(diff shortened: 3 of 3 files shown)


test/README.md
```diff
@@ -7,35 +7,59 @@ This folder contains a test application (`fastrpc_test.c`) that demonstrates the
 After building the application, the following files and directories need to be pushed to the device:
 
 1. **fastrpc_test Binary**: The compiled test application.
-2. **android Directory**: Contains shared libraries for the Android platform.
-3. **linux Directory**: Contains shared libraries for the Linux platform.
+2. **linux Directory**: Contains shared libraries for the Linux platform.
+3. **android Directory**: Contains shared libraries for the Android platform.
```
lumag: Why did you swap them?

test/README.md
```diff
@@ -7,35 +7,59 @@ This folder contains a test application (`fastrpc_test.c`) that demonstrates the
...
+### For Linux
 
-- `/path/to/your/fastrpc/test/.libs/fastrpc_test`
-- `/path/to/your/fastrpc/test/android` (if using Android)
-- `/path/to/your/fastrpc/test/linux` (if using Linux)
+Use `make install` from the `test` directory to install the libraries and the executable to the appropriate directories:
+
+```bash
+cd /path/to/your/fastrpc/test
+make install
```
lumag: Why is it necessary to change to a subdir? Installation is a recursive target. Not to mention that the base should be to run  tests from the build dir without installation.

test/README.md
```diff
@@ -7,35 +7,59 @@ This folder contains a test application (`fastrpc_test.c`) that demonstrates the
...
+1. Navigate to the `/usr/local/bin` directory on the device.
 2. Execute the `fastrpc_test` binary with the appropriate options.
 
 Example command:
 
 ```bash
-cd /usr/bin
+cd /usr/local/bin
+chmod 777 fastrpc_test
 ./fastrpc_test -d 3 -U 1 -t linux -a v68
```
lumag: Yuck. No. 777 is a definite mistake. And make can do this for you anyway.

test/README.md
```diff
@@ -7,35 +7,59 @@ This folder contains a test application (`fastrpc_test.c`) that demonstrates the
...
+### For Android
+
+1. Navigate to the custom path on the device.
+2. Execute the `fastrpc_test` binary with the appropriate options.
+
+Example command:
+
+```bash
+cd /vendor/lib/rfsa/dsp/sdk/
+chmod 777 fastrpc_test
```
lumag: Same here. Don't make system files writable to everybody.

test/README.md
```diff
@@ -7,35 +7,59 @@ This folder contains a test application (`fastrpc_test.c`) that demonstrates the
...
+./fastrpc_test -d 3 -U 1 -t android -a v68
```
lumag: Also... Why do you need to cd at all? Just rub the command from the system path.

test/README.md
quic-vkatoch: On Android, certain custom paths (like /vendor) are used specifically for testing. To document that use case, added limited information about it. Keeping Linux-specific section first, as it covers the more general scenario.

test/README.md
quic-vkatoch: Ack, no need to change to the test subdir.

test/README.md
quic-vkatoch: Ack.

test/README.md
quic-vkatoch: Ack.

test/README.md
quic-vkatoch: Ack.

ekanshibu (dismissed):

test/README.md
lumag: Don't add extra noise to the patch. If you want to reorganise something, split it to a separate patch.

test/fastrpc_test.c
```diff
@@ -75,44 +74,49 @@ int main(int argc, char *argv[]) {
...
-    // Construct the absolute library path
-    snprintf(abs_lib_path, sizeof(abs_lib_path), "%s", target);
-
-    if (realpath(abs_lib_path, abs_lib_path) == NULL) {
-        fprintf(stderr, "Error resolving path %s: %s\n", abs_lib_path, strerror(errno));
-        return -1;
+    // Construct the library paths
+    if (strcmp(target, "linux") == 0) {
+        snprintf(ld_lib_path, sizeof(ld_lib_path), "%s", testlibdir);
+        snprintf(dsp_lib_path, sizeof(dsp_lib_path), "%s/%s", testdspdir, arch_version);
```
lumag: Those environment variables might have values which are required by the user. Please change the code to extend them instead of overriding the values.

test/fastrpc_test.c
```diff
@@ -75,44 +74,49 @@ int main(int argc, char *argv[]) {
...
+    } else if(strcmp(target, "android") == 0) {
+        snprintf(ld_lib_path, sizeof(ld_lib_path), "%s", target);
```
lumag: Why?

test/fastrpc_test.c
```diff
@@ -75,44 +74,49 @@ int main(int argc, char *argv[]) {
...
+    else {
+        printf("\nERROR: Invalid target platform (-t). Must be linux or android.\n");
+        print_usage();
         return -1;
     }
 
-    // Construct LD_LIBRARY_PATH and DSP_LIBRARY_PATH
-    snprintf(ld_lib_path, sizeof(ld_lib_path), "%s", abs_lib_path);
-
+    // Set the new LD_LIBRARY_PATH
```
lumag: Don't add useless obvious comments

lumag (changes requested):
Commit message is incorrect. You are not just 'updating readme'. You are changing how the code works.

test/README.md
quic-vkatoch: Ack.

test/fastrpc_test.c
quic-vkatoch: ld_lib_path and dsp_lib_path are local variables used for internal path construction. They do not directly override environment variables.

test/fastrpc_test.c
quic-vkatoch: Will modify it to same as it is for linux.

test/fastrpc_test.c
quic-vkatoch: Sure, I'll keep that in mind.

test/fastrpc_test.c
lumag: Yes. But then you use them inside setenv() call. I don't see your point.

quic-vkatoch: Updated to append these paths.

test/Makefile.am
```diff
@@ -5,7 +5,9 @@ bin_PROGRAMS = fastrpc_test
 fastrpc_test_SOURCES = fastrpc_test.c
 
 # Define the compiler flags for the test program
```
ekanshibu, line 7: write better commit messages

quic-vkatoch: Updated

PR #202, opened 2025-07-21
src, inc: fix CRLF file endings

lumag:
Files under Git control should not be using CR-LF for EOL. Converting to native EOL symbols is handled by the git itself. Remove extra CR-LF terminators.


quic-ekangupt:
This change looks to be already part of PR#203
https://github.com/qualcomm/fastrpc/pull/203/commits/f1f04fc5ec564e6b77613c1959af7a8993bff19c

Chennak-quic (commented):
rebase required.

PR #204, opened 2025-07-23
Fix warning: variable might be uninitialized

quic-jianpl:
Move the definition of args to the beginning of the function(remote_handle_invoke_domain), which seems to solve the variable might be uninitialized warning.
Since there is a goto bail later in the function, the compiler might think that args could still be used without a definition.

### DIFF
--- a/inc/fastrpc_ioctl.h
+++ b/inc/fastrpc_ioctl.h
@@ -57,8 +57,7 @@
 	#define DEFAULT_DEVICE ""
 #endif
 
-#define INITIALIZE_REMOTE_ARGS(total)	struct fastrpc_invoke_args* args = NULL; \
-					int *pfds = NULL; \
+#define INITIALIZE_REMOTE_ARGS(total)	int *pfds = NULL; \
 					unsigned *pattrs = NULL; \
 					args = (struct fastrpc_invoke_args*) calloc(sizeof(*args), total);	\
 					if(args==NULL) { 	\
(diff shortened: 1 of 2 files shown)


inc/fastrpc_ioctl.h
```diff
@@ -57,8 +57,7 @@
 	#define DEFAULT_DEVICE ""
 #endif
 
-#define INITIALIZE_REMOTE_ARGS(total)	struct fastrpc_invoke_args* args = NULL; \
-					int *pfds = NULL; \
+#define INITIALIZE_REMOTE_ARGS(total)	int *pfds = NULL; \
```
lumag, line 60: Plese fix the commit and commit message:
- Use your actual name in Git config, not your nick name / login name
- Add proper commit message.

lumag (changes requested):

quic-ekangupt:
is this commit really needed:
https://github.com/qualcomm/fastrpc/pull/204/commits/9e85d0a22a9f31f61cf60c5effc6c7e4c08b2a77

Chennak-quic (approved):

src/fastrpc_apps_user.c
```diff
@@ -1168,6 +1168,7 @@ int remote_handle_invoke_domain(int domain, remote_handle handle,
   fastrpc_timer frpc_timer;
   int trace_marker_fd = hlist[domain].trace_marker_fd;
   bool trace_enabled = false;
+  struct fastrpc_invoke_args* args = NULL;
```
ekanshibu, line 1171: can you keep 1 empty line before Signed-off-by:?

inc/fastrpc_ioctl.h
ekanshibu: Hope your concerns are resolved, planning to merge this PR.

ekanshibu: Marking the conversation as resolved.

PR #210, opened 2025-08-06
Remove fallback to default device open

quic-lxu5:
In case of default domain ID, the device open request falls back to DEFAUL_DEVICE which will be defined in the libs to which the application is linked. Remove this legacy fallback logic to avoid any confusion and unexpected observations.


quic-ekangupt:
Can you also clean-up the following:

1. This function: https://github.com/qualcomm/fastrpc/blob/8c8e7e050a4bde06ea75623115f2c6e108b1a323/src/adsp_default_listener.c#L95
2. DEFAULT_DEVICE from here:
https://github.com/qualcomm/fastrpc/blob/8c8e7e050a4bde06ea75623115f2c6e108b1a323/src/fastrpc_ioctl.c#L44
https://github.com/qualcomm/fastrpc/blob/8c8e7e050a4bde06ea75623115f2c6e108b1a323/inc/fastrpc_ioctl.h#L45
https://github.com/qualcomm/fastrpc/blob/8c8e7e050a4bde06ea75623115f2c6e108b1a323/src/fastrpc_apps_user.c#L3264

src/adsp_default_listener.c
```diff
@@ -85,35 +85,7 @@ static const char *get_secure_device_name(int domain_id) {
...
 		break;
 	default:
-		name = DEFAULT_DEVICE;
-		break;
-	}
-
-	return name;
-}
-
-static const char *get_default_device_name(int domain_id) {
```
ekanshibu, line 107: okay, I checked and this function cannot be removed. The reason is that some old platform use non-secure devices for daemons(if non-secure-domain property is added in DT).

You can just rename the function as get_non_secure_device_name and modify the default: case here.

src/adsp_default_listener.c
```diff
@@ -154,13 +126,12 @@ static bool fastrpc_dev_exists(const char* dev_name)
 static int fastrpc_wait_for_device(int domain)
 {
 	int inotify_fd = -1, watch_fd = -1, err = 0;
-	const char *sec_dev_name = NULL, *def_dev_name = NULL;
```
ekanshibu, line 175: check for non-secure device also to maintain backward compatibility

quic-vkatoch:
Closing this PR, it has been superseded by PR #302.

PR #213, opened 2025-08-11
clang-format

quic-kdybcio:
Loosely based on & supersedes #136 

The config file generally resembles the code style one would expect in the kernel source tree

The last commit is a result of manually running the command that the build workflow executes and can be verified independently, the clang-format version I ran is precisely `Ubuntu clang-format version 20.1.8 (++20250804090239+87f0227cb601-1~exp1~20250804210352.139)`


quic-kdybcio:
On projects using GNU make, it's commonplace to add a `make format` target, I don't know how to integrate that with automake but it'd probably be useful too..

quic-kdybcio:
Pushed a change to remove spaces before braces (which we clearly don't want)

Chennak-quic:
Please resolve merge conflicts

PR #216, opened 2025-08-14
rpcmem: Register rpcmem-allocated buffers with fastrpc memory framework

quic-parihar:
Currently, rpcmem-allocated buffers used in remote calls are not recognized by the RPC kernel driver, resulting in fallback to copy buffers for message passing. This fallback introduces significant performance degradation, which worsens with increasing buffer sizes.

This commit ensures that rpcmem-allocated buffers are properly registered with the fastrpc memory framework upon allocation and deregistered upon release. By enabling the kernel driver to identify the associated file descriptors, it avoids the use of copy buffers during remote invocation, thereby improving performance and efficiency.

### DIFF
--- a/inc/fastrpc_mem.h
+++ b/inc/fastrpc_mem.h
@@ -55,4 +55,22 @@ int remote_mmap64_internal(int fd, uint32_t flags, uint64_t vaddrin, int64_t siz
 */
 int fastrpc_buffer_ref(int domain, int fd, int ref, void **va, size_t *size);
 
+
+/*
+ * Register or deregister a buffer with fastrpc framework, managing its reference count
+ * and associated metadata.
+ *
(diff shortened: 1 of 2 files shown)


inc/fastrpc_mem.h
```diff
@@ -55,4 +55,22 @@ int remote_mmap64_internal(int fd, uint32_t flags, uint64_t vaddrin, int64_t siz
 */
 int fastrpc_buffer_ref(int domain, int fd, int ref, void **va, size_t *size);
 
+
+/*
+ * Register or deregister a buffer with the remote subsystem, managing its reference count
```
ekanshibu: Register or deregister a buffer with **fastrpc framework**

quic-parihar: Ack

Chennak-quic (approved):

PR #234, opened 2025-09-04
Update DSP mount and search path configuration

quic-parihar:
Update DSP mount and search path configuration to
support dynamic runtime resolution
- Refactored mount logic to allow flexible path overrides
- Added support for environment-based search path injection


files/machine_dsp_paths.json
```diff
@@ -0,0 +1,10 @@
+{
+    "Qualcomm Technologies, Inc. DB820c": "/usr/share/qcom/apq8096/Qualcomm/db820c/",
+    "Thundercomm Dragonboard 845c": "/usr/share/qcom/sdm845/Thundercomm/db845c/",
+    "Qualcomm Technologies, Inc. Robotics RB5": "/usr/share/qcom/sm8250/Thundercomm/RB5/",
+    "Qualcomm Technologies, Inc. Robotics RB1": "/usr/share/qcom/qcm2290/Thundercomm/RB1/",
+    "Qualcomm Technologies, Inc. QRB4210 RB2": "/usr/share/qcom/qrb4210/Thundercomm/RB2/",
+    "Qualcomm Technologies, Inc. Robotics RB3gen2": "/usr/share/qcom/qcm6490/Thundercomm/RB3gen2/",
+    "Qualcomm QCS9100 Ride Rev3": "/usr/share/qcom/sa8775p/Qualcomm/SA8775P-RIDE/",
+    "Qualcomm Technologies, Inc. QCS8300 Ride":"/usr/share/qcom/qcs8300/Qualcomm/QCS8300-RIDE/"
```
lumag: Can we please have some more flexible config file in a format more useful than JSON?
E.g. we might want to add a property whether the device has separate SDSP or whether sensorspd should be handled by the ADSP too.
I'd recommend using libconfig and ini format for the config file.

Makefile.am
```diff
@@ -1,3 +1,3 @@
-SUBDIRS = inc src test
+SUBDIRS = inc src test files
```
ekanshibu: use qti mail id

inc/apps_std_internal.h
```diff
@@ -22,35 +22,13 @@
...
-#define DSP_DOM_LOCATION "/usr/lib/dsp/xdspn"
-#endif
-#endif /* ENABLE_UPSTREAM_DRIVER_INTERFACE */
-
 #ifndef VENDOR_DSP_LOCATION
 #define VENDOR_DSP_LOCATION "/vendor/dsp/"
 #endif
 #ifndef VENDOR_DOM_LOCATION
 #define VENDOR_DOM_LOCATION "/vendor/dsp/xdsp/"
 #endif
```
ekanshibu, line 31: @quic-vkatoch, pls check if this aligns with your plan to drop vendor/android specific paths

quic-vkatoch: Sure

files/machine_dsp_paths.json
quic-parihar: Reviewing this, Planning to move ahead with ini file type

Makefile.am
quic-parihar: Acked

lumag:
https://github.com/linux-msm/hexagonrpc/issues/15 triggered the following idea: the binaries can be coming from different sources. So we might not have a single .conf file. We need to be able to parse `conf.d`-like directory, each file adding new entries and new machines. This way hexagon-dsp-binaries can provide configuration file for the binaries inside that package, PmOS can install additional device-specific binaries and then the user can manually repack Hexagon binaries for the WoA laptops.

ekanshibu:
> [linux-msm/hexagonrpc#15](https://github.com/linux-msm/hexagonrpc/issues/15) triggered the following idea: the binaries can be coming from different sources. So we might not have a single .conf file. We need to be able to parse `conf.d`-like directory, each file adding new entries and new machines.

I see, if the sources of binaries are going to change, then yes, a single file might not work. I think this modular configuration also covers the requirement of supporting vendor specific DSP artifacts. I don't have the complete understanding on `conf.d` but will check this. @quic-parihar you can also have a look into this.

> This way hexagon-dsp-binaries can provide configuration file for the binaries inside that package, PmOS can install additional device-specific binaries and then the user can manually repack Hexagon binaries for the WoA laptops.

I don't completely understand the manually repacking point. Is PmOS going to maintain the device specific binaries separately? Or the bins are still maintained in hexagon-dsp-binaries and the user is somehow making the modification to use different binaries?

@quic-bkumar @Chennak-quic, seeking your inputs as well.

lumag:
> > > Thinking that instead of reading all the conf files from conf.d directory for configuring dsp bin paths, shouldn't we have a centralized place like we have for linux dsp firmware binaries which tells us where to read the bins from. In this case I was thinking a file or environment variable which tells which conf/ini file to refer for configuring dsp bin paths. @lumag need your inputs on this.
> > 
> > 
> > Having an environment variable would defeat the purpose of conf.d. You have to read all config files and select the correct entry.
> 
> If there are two .conf files—one provided by Qualcomm and the other uploaded by the user—and both contain a machine entry for "Qualcomm QCS9100 Ride Rev3" defining different paths, how can we determine which file specifies the correct path for that entry?

Let user decide. Introduce some logical order (e.g. alphabetical) and then the later definition "wins".

src/fastrpc_apps_user.c
```diff
@@ -4157,6 +4147,79 @@ static void exit_thread(void *value) {
...
+        strlcat(dsp_dom, ";", sizeof(dsp_dom));
+        strlcat(dsp_dom, token, sizeof(dsp_dom));
+        strlcat(dsp_dom, suffix, sizeof(dsp_dom));
+        token = strtok(NULL, ";");
+    }
+    strlcat(dsp_dom, ";", sizeof(dsp_dom));
+    setenv("DSP_DOMAIN_LOCATION", dsp_dom, 1);
+}
+
+static void update_dsp_search_path(char *buffer) {
```
kadamski: The argument could be "const char *"

src/fastrpc_apps_user.c
```diff
@@ -4157,6 +4147,79 @@ static void exit_thread(void *value) {
...
+  char dsp_search_paths[MAX_PATH_LEN];
+
+  strlcpy(dsp_search_paths, getenv("DSP_SEARCH_PATH"), sizeof(dsp_search_paths));
+  strlcat(dsp_search_paths, buffer, sizeof(dsp_search_paths));
+  strlcat(dsp_search_paths, ";", sizeof(dsp_search_paths));
+  setenv("DSP_SEARCH_PATH", dsp_search_paths, 1);
+}
+
+static void update_dsp_mount_path(char *buffer) {
```
kadamski: The argument could be "const char *"

src/fastrpc_apps_user.c
```diff
@@ -76,16 +76,15 @@
...
-#define DSP_MOUNT_LOCATION "/usr/lib/dsp/"
-#define DSP_DOM_LOCATION "/usr/lib/dsp/xdspn"
-#endif
+#define MACHINE_DSP_FILE_PATH "/usr/share/qcom/machine_dsp_paths.ini"
+#define DEFAULT_DSP_MOUNT_LOCATION ";/usr/lib/dsp/;"
+#define DEFAULT_DSP_SEARCH_PATH ";/usr/lib/rfsa/adsp;/vendor/lib/rfsa/adsp;/vendor/dsp/;/usr/lib/dsp/;"
 #define VENDOR_DSP_LOCATION "/vendor/dsp/"
 #define VENDOR_DOM_LOCATION "/vendor/dsp/xdsp/"
 
+#define MAX_PATH_LEN 256
```
kadamski: I'm just wondering why you use arbitrary 256 here? Why not something like PATH_MAX?

As a side note, is it a good idea to silently ignore the situation when the paths are truncated because the buffer was too short? This define is used in some cases to allocate the buffer that will hold multiple paths, not just one, and 256 seems a bit small already for a single path.

src/fastrpc_apps_user.c
```diff
@@ -4157,6 +4147,79 @@ static void exit_thread(void *value) {
   pthread_setspecific(tlsKey, (void *)NULL);
 }
 
+const char* get_dsp_search_path() {
+  return getenv("DSP_SEARCH_PATH");
+}
+
+void update_dsp_dom_location(const char *suffix) {
```
kadamski: Shouldn't be static?

src/log_config.c
```diff
@@ -416,7 +417,7 @@ static void *file_watcher_thread(void *arg) {
       errno = current_errno;
       // User has not set the env variable. Get default search paths.
```
kadamski: I think this comment is not valid anymore. You are now using the search paths from the env as a fallback now and this may have been set by the user via the .ini file. So this is no longer about using statically defined default search paths.

src/log_config.c
```diff
@@ -416,7 +417,7 @@ static void *file_watcher_thread(void *arg) {
...
       if (ret != 0)
-        memmove(data_paths, DSP_SEARCH_PATH, strlen(DSP_SEARCH_PATH));
+        memmove(data_paths, dsp_search_path, strlen(dsp_search_path));
```
kadamski: So dsp_search_path is a string representing the environment variable DSP_SEARCH_PATH that could maybe be now be set externally so assuming it will for sure fit into data_paths seems a bit risky. I think using strlcpy here with ENV_PATH_LEN as a limit would be safer.

src/fastrpc_apps_user.c
```diff
@@ -4157,6 +4147,79 @@ static void exit_thread(void *value) {
...
+
+  while (fgets(line, sizeof(line), file)) {
+    if (sscanf(line, "[%[^]]]", section) == 1) {
+      if (section_found)
+        break;
+      section_found = strcmp(section, buffer) == 0;
+    } else if (section_found && sscanf(line, " %[^=]= \"%[^\"]\"", key, value) == 2) {
+      if (strcmp(key, "path") == 0)
+        break;
+    }
```
kadamski: Shouldn't there be some kind of a warning or an error in case the file has a syntax error and the sscanf directive doesn't match? Like if someone would forget to put quote in the value? I think it might be hard to debug such a case otherwise.

quic-parihar:
> > > > Thinking that instead of reading all the conf files from conf.d directory for configuring dsp bin paths, shouldn't we have a centralized place like we have for linux dsp firmware binaries which tells us where to read the bins from. In this case I was thinking a file or environment variable which tells which conf/ini file to refer for configuring dsp bin paths. @lumag need your inputs on this.
> > > 
> > > 
> > > Having an environment variable would defeat the purpose of conf.d. You have to read all config files and select the correct entry.
> > 
> > 
> > If there are two .conf files—one provided by Qualcomm and the other uploaded by the user—and both contain a machine entry for "Qualcomm QCS9100 Ride Rev3" defining different paths, how can we determine which file specifies the correct path for that entry?
> 
> Let user decide. Introduce some logical order (e.g. alphabetical) and then the later definition "wins".

@lumag I understand the purpose of the conf.d directory — we're using it to manage multiple configuration files for different purposes.

However, I'm still unclear about the need to read multiple .ini files specifically for configuring dsp search paths for loading firmware libs. If a user wants to define custom paths for firmware libraries, can't they simply modify the existing machine_dsp_paths.ini file or replace it with a new file of the same name?
The only challenge I see is reverting to the default paths defined in the original .ini file. But even that can be handled by keeping a backup of the original before making changes.
Introducing multiple files for dsp search paths and relying on users to maintain them in a particular order might lead to unnecessary complexity and confusion.

Let me know what I am missing to align on same page with you.

files/machine_dsp_paths.ini
Chennak-quic, line 1: I think this config file should be created by hexagon-dspso packaging recipe (https://github.com/qualcomm-linux/meta-qcom/tree/master/recipes-bsp/hexagon-dspso).

lumag: those recipes will create config files for the contents of those packages.

quic-parihar:
Based on the internal and discussion we have had on this PR, proposing the following:
We expect packages to bring in their own config files whether it be hexagon-dsp-binaries or some other package.
Will be parsing /usr/share/qcom/conf.d/ directory to read all the .conf files present in lexicographical order. Picking the latest dsp_lib_path.

config file property will be DSP_library_path=<path_value>

For backward compatibility will also maintain a default config file with dsp_lib_paths configured in fastrpc project.

src/fastrpc_apps_user.c
quic-parihar: acked

src/fastrpc_apps_user.c
quic-parihar: acked

src/fastrpc_apps_user.c
quic-parihar: Going with PATH_MAX

src/fastrpc_apps_user.c
quic-parihar: acked

src/log_config.c
quic-parihar: acked, thanks for pointing this out

src/log_config.c
quic-parihar: acked, thanks for pointing this out

Docs/conf_guideline.md
```diff
@@ -0,0 +1,26 @@
...
+Do not modify section headers ([Platform Name]) unless adding a new supported platform.
+Ensure DSP_library_path values are enclosed in double quotes ("...") and are absolute.
+Avoid trailing spaces or special characters that may break parsing.
+Maintain consistent indentation and line breaks.
+Add new property under same platform name, don't create new duplicate platform entries.
+When adding new properties document their purpose in Readme file.
+➕ Adding New Platforms
+To add a new platform, follow the existing format:
+[New Platform Name]
+DSP_library_path = "/usr/share/qcom/new_platform/path/"
```
lumag: I thought that you've agree to a comment here that paths should be relative. Any reason for the change?

src/fastrpc_apps_user.c
```diff
@@ -76,16 +76,16 @@
...
-#define DSP_MOUNT_LOCATION "/dsp/"
-#define DSP_DOM_LOCATION "/dsp/xdsp"
-#else
-#define DSP_MOUNT_LOCATION "/usr/lib/dsp/"
-#define DSP_DOM_LOCATION "/usr/lib/dsp/xdspn"
+#ifndef CONFIG_DIR
+#define CONFIG_DIR "/usr/share/qcom/conf.d/"
 #endif
+#define DEFAULT_DSP_MOUNT_LOCATION ";/usr/lib/dsp/;"
+#define DEFAULT_DSP_SEARCH_PATH ";/usr/lib/rfsa/adsp;/vendor/lib/rfsa/adsp;/vendor/dsp/;/usr/lib/dsp/;"
```
lumag: These locations should go and never be used again. None of them confirms to LInux FHS.

inc/apps_std_internal.h
```diff
@@ -22,35 +22,13 @@
...
-#ifndef DSP_DOM_LOCATION
-#define DSP_DOM_LOCATION "/usr/lib/dsp/xdspn"
-#endif
-#endif /* ENABLE_UPSTREAM_DRIVER_INTERFACE */
-
 #ifndef VENDOR_DSP_LOCATION
 #define VENDOR_DSP_LOCATION "/vendor/dsp/"
 #endif
 #ifndef VENDOR_DOM_LOCATION
 #define VENDOR_DOM_LOCATION "/vendor/dsp/xdsp/"
```
lumag, line 29: These should go too

src/fastrpc_apps_user.c
```diff
@@ -3462,23 +3464,12 @@ static int open_shell(int domain_id, apps_std_FILE *fh, int unsigned_shell) {
...
   strlcat(shell_absName, domain_str, shell_absNameLen);
 
-  absNameLen = strlen(DSP_MOUNT_LOCATION) + shell_absNameLen + 1;
-  VERIFYC(NULL != (absName = (char *)malloc(sizeof(char) * absNameLen)),
-          AEE_ENOMEMORY);
-  strlcpy(absName, DSP_MOUNT_LOCATION, absNameLen);
-  strlcat(absName, shell_absName, absNameLen);
-
-  nErr = apps_std_fopen(absName, "r", fh);
+  nErr = apps_std_fopen_with_env("DSP_MOUNT_LOCATION", ";", shell_absName, "r", fh);
```
lumag: Why do you still need DSP_MOUNT_LOCATION handling?

src/fastrpc_apps_user.c
```diff
@@ -4157,6 +4148,159 @@ static void exit_thread(void *value) {
...
+    free(file_list[i]);
+  }
+
+  const char *resolved_path = resolve_to_abs_path(dsp_lib_paths);
+  if (!resolved_path)
+    return;
+
+  strlcpy(dsp_mount_location, resolved_path, sizeof(dsp_mount_location));
+  strlcat(dsp_mount_location, getenv("DSP_MOUNT_LOCATION"), sizeof(dsp_mount_location));
+  setenv("DSP_MOUNT_LOCATION", dsp_mount_location, 1);
```
lumag: Please stop playing with environment variables. Just use a normal global variable for that.

inc/apps_std_internal.h
quic-parihar: This is being addressed in [PR #154](https://github.com/qualcomm/fastrpc/pull/154/files). @quic-vkatoch has begun reworking the implementation and will clean up this area along with a few additional components.

src/fastrpc_apps_user.c
quic-parihar: This is being addressed in [PR #154](https://github.com/qualcomm/fastrpc/pull/154/files). @quic-vkatoch has begun reworking the implementation and will clean up this area along with a few additional components.

Docs/conf_guideline.md
quic-parihar: Yes we are going ahead with relative and absolute paths. Based on internal discussion have decided to leave it to user. Will update same in .md
Relative will be relative to current path i.e. /usr/share/qcom/conf.d

Example Entries
[Qualcomm Technologies, Inc. DB820c]
DSP_library_path = "/usr/share/qcom/apq8096/Qualcomm/db820c/"
DSP_library_path = "../apq8096/Qualcomm/db820c/"

lumag: ack

lumag: Maybe it should be either absolute or relative to the "root" - i.e. `/usr/share/qcom` path. This will remove the need for an extra `../` in it.

src/fastrpc_apps_user.c
quic-parihar: Ack

src/fastrpc_apps_user.c
quic-parihar: DSP_MOUNT_LOCATION was pointing to dsp lib paths, I went with old naming convention. Renaming it to DSP_LIBRARY_PATH.

src/fastrpc_apps_user.c
quic-parihar: Ack

quic-parihar:
> The code needs to be cleaned up to get rid of handling everything through environment variables.

Acked

ekanshibu (dismissed):

src/fastrpc_apps_user.c
```diff
@@ -4157,6 +4153,100 @@ static void exit_thread(void *value) {
...
+    free(file_list[i]);
+  }
+
+  strlcpy(DSP_LIBS_LOCATION, RELATIVE_LIB_PATH, sizeof(DSP_LIBS_LOCATION));
+  strlcat(DSP_LIBS_LOCATION, dsp_lib_paths, sizeof(DSP_LIBS_LOCATION));
+  update_dsp_search_path(DSP_LIBS_LOCATION);
+}
+
+static void configure_dsp_paths() {
+  char buffer[PATH_MAX] = {0};
```
Chennak-quic: as you are reading machine name into buffer, can we rename this variable to make it more readable ?

src/fastrpc_apps_user.c
```diff
@@ -4157,6 +4153,100 @@ static void exit_thread(void *value) {
...
+  FILE *file = fopen("/sys/firmware/devicetree/base/model", "r");
```
Chennak-quic: This path can be platform specific. can we configure this in platform-specific make files ?

src/fastrpc_apps_user.c
```diff
@@ -4157,6 +4153,100 @@ static void exit_thread(void *value) {
...
+    FARF(ALWAYS, "Warning: DSP_library_path directive missing/empty or doesn't "
+          "match expected syntax in section [%s].\n", buffer);
+    return;
+  }
+
+  strlcpy(dsp_lib_paths, value, sizeof(value));
+}
+
+static void parse_config_dir(char *buffer) {
+  DIR *dir = opendir(CONFIG_DIR);
```
Chennak-quic: CONFIG_DIR is not documented.

src/fastrpc_apps_user.c
```diff
@@ -4157,6 +4153,100 @@ static void exit_thread(void *value) {
...
+  if (!file)
+    return;
+
+  while (fgets(line, sizeof(line), file)) {
+    if (sscanf(line, "[%[^]]]", section) == 1) {
+      if (section_found)
+        break;
+      section_found = strcmp(section, buffer) == 0;
+    } else if (section_found && sscanf(line, " %[^=]= \"%[^\"]\"", key, value) == 2) {
+      if (strcmp(key, "DSP_library_path") == 0)
```
Chennak-quic: can we define this key as macro ?

Docs/conf_guideline.md
```diff
@@ -0,0 +1,26 @@
+📄 conf Usage Guide
+
+🔧 Purpose
+The <>.conf file provides platform-specific paths used by automation scripts and system tools to locate DSP binaries and resources. Each section corresponds to a specific hardware platform.
+📁 Format Guidelines
+Each entry in the .conf file must follow this structure:
+[Platform Name]
+DSP_library_path = "<relative path to DSP binaries and resources for the platform>"
+✅ Example Entries
+[Qualcomm Technologies, Inc. DB820c]
```
Chennak-quic: As a user how do i get the machine name for my platform ? 
How does fastrpc find out machine name at run time ?

Docs/conf_guideline.md
```diff
@@ -0,0 +1,26 @@
+📄 conf Usage Guide
+
+🔧 Purpose
+The <>.conf file provides platform-specific paths used by automation scripts and system tools to locate DSP binaries and resources. Each section corresponds to a specific hardware platform.
```
Chennak-quic: Can you please elaborate on why fastrpc framework needs this conf file. 
I think property name can DSP_LIBRARY_PATH to make the name same as the environment variable that already exposed to Hexagon SDK clients.

src/fastrpc_apps_user.c
quic-parihar: Ack

src/fastrpc_apps_user.c
quic-parihar: Acked

Docs/conf_guideline.md
quic-parihar: Acked

Docs/conf_guideline.md
quic-parihar: Acked

src/fastrpc_apps_user.c
quic-parihar: Acked

src/fastrpc_apps_user.c
quic-parihar: Acked

ekanshibu (dismissed):

quic-parihar:
> The code needs to be cleaned up to get rid of handling everything through environment variables.

Acked

Chennak-quic (approved):

Docs/conf_guideline.md
```diff
@@ -0,0 +1,65 @@
+📄 **conf Usage Guide**
+
+---
+
+### 🔧 **Purpose**
+The `<>.conf` file enables **fastrpc** to set machine-specific configurations at runtime. Each section corresponds to a specific hardware platform.
+
+- fastrpc supports reading config files from a particular directory. Users should ensure all configuration files are stored in that same directory.
```
lumag: Please specify what happens if there are multiple config files defining a path for a single machine.

src/log_config.c
```diff
@@ -414,9 +416,8 @@ static void *file_watcher_thread(void *arg) {
       ret = apps_std_getenv(DSP_LIBRARY_PATH, data_paths, ENV_PATH_LEN,
                             &env_list_len);
       errno = current_errno;
-      // User has not set the env variable. Get default search paths.
       if (ret != 0)
-        memmove(data_paths, DSP_SEARCH_PATH, strlen(DSP_SEARCH_PATH));
+        strlcpy(data_paths, dsp_search_path, ENV_PATH_LEN);
```
lumag, line 420: Why do we need a separate search path here? What is the difference between `DSP_SEARCH_PATH` and `DSP_LIBRARY_PATH`?

src/fastrpc_apps_user.c
```diff
@@ -4157,6 +4154,100 @@ static void exit_thread(void *value) {
...
+  FILE *file = fopen(filepath, "r");
+  if (!file)
+    return;
+
+  while (fgets(line, sizeof(line), file)) {
+    if (sscanf(line, "[%[^]]]", section) == 1) {
+      if (section_found)
+        break;
+      section_found = strcmp(section, machine_name) == 0;
+    } else if (section_found && sscanf(line, " %[^=]= \"%[^\"]\"", key, value) == 2) {
```
lumag: Can we _please_ use libconfig or some other existing library instead of parsing it manually? This code would be impossible to expand once we need to add more values to the configuration file.

quic-vkatoch: I'm exploring libconfig, and I see that we need to install the library before compiling fastrpc. Is that fine, or is there an alternative?

Docs/conf_guideline.md
```diff
@@ -0,0 +1,65 @@
...
+DSP_LIBRARY_PATH = "/sdm845/Thundercomm/db845c/"
+```
+
+---
+
+### ⚠️ **Important Notes**
+- Do **not** modify section headers (`[Machine Name]`) unless adding a new supported Machine.
+- Ensure `DSP_LIBRARY_PATH` values:
+  - Are enclosed in double quotes (`"..."`).
+  - Are **relative to `/usr/share/qcom/`**.
```
basak-qcom, line 54: `/usr/share/qcom/` isn't acceptable to Debian. `/usr/share/<purpose>/` is expected, not `/usr/share/<vendor>/`, so for hexagon-dsp-binaries I have insisted against `/usr/share/qcom/` for hexagon-dsp-binaries in favour of `/usr/share/hexagon-dsp/`.

If you hardcode `/usr/share/qcom/`, then the Debian package will need to patch this, which we should try to avoid.

@lumag do you think meta-qcom will switch to `/usr/share/hexagon-dsp/`, or will this need to be configurable?

Docs/conf_guideline.md
```diff
@@ -0,0 +1,65 @@
+📄 **conf Usage Guide**
+
+---
+
+### 🔧 **Purpose**
+The `<>.conf` file enables **fastrpc** to set machine-specific configurations at runtime. Each section corresponds to a specific hardware platform.
+
+- fastrpc supports reading config files from a particular directory. Users should ensure all configuration files are stored in that same directory.
+  - For Linux platforms: `/usr/share/qcom/conf.d/`
```
basak-qcom, line 9: As below, `/usr/share/<vendor>/` is not acceptable to Debian. `/usr/share/fastrpc/` is preferable. Or, if this file is intended to be generic to all consumers of hexagon-dsp, then it should be shipped by hexagon-dsp-binaries instead, under `/usr/share/hexagon-dsp/`.

You might consider the file to be "global Qualcomm hardware information" and have keys apart from `DSP_LIBRARY_PATH` in the future. However this isn't conducive to per-package distribution packaging for non-image-based distributions and I think we should avoid that. I can go into details if you'd like.

src/fastrpc_apps_user.c
lumag: What is the problem with installing libconfig?

basak-qcom: I'm not familiar with libconfig myself, but it does sound like the right sort of tool to use here. I see that it is already packaged in Debian. In the upstream ecosystem it is completely normal to depend on libraries also available in that ecosystem, and for the developer to be expected to have those libraries installed and ready before attempting to build a project that depends on them. In general, if it's a regular C library packaged in Debian, then from an ecosystem and upstream quality perspective it's preferable to use it over attempting to reinvent its functionality. It is preferable to dynamically link to such things over embedding them into the source tree.

quic-vkatoch: Ack

Docs/conf_guideline.md
lumag: I think, it is already configurable at the build time, so Debian will use a different setting here. I'd prefer to describe `/usr/share/qcom` here (as it is the path that has been agreed between several projects) and explicitly state that other distributions might use different directory here.

Docs/conf_guideline.md
lumag: Let's not multiply instances. `/usr/share/qcom/conf.d` by default, `/usr/share/qcom/hexagon-dsp/conf.d` on Debian.

lumag:
> Is this the right approach? Since we’re new to this, could you advise?

Of course not. What stops you from `apt install libyaml-dev` and then simply building FastRPC as a normal software on your laptop?

BTW: libconfig is going to be added to OE-Core in a few days, so if you want to use it, the restrictions are now lifted

quic-vkatoch:
Sure, I’ll experiment with this. For standalone compilation, I also need to check how this dependency behaves under cross-compilation (host: x86, target: aarch64). I’m seeking the right way to make it work for cross-compilation—could you advise? 

I also successfully built these changes with meta‑qcom by adding a libyaml dependency in the .bb file.

src/log_config.c
quic-parihar: Was following older approach with 2 different path definitions maintained for shell and dsp libs. Have moved to a single path as I see no requirement of maintaining them separately.

basak-qcom:
Sorry, I think I've expressed this somewhere before but perhaps not in this PR.

Why do we need a search path at all? I understand that paths are different depending on the build target. But that can be resolved at build time. At runtime, why is it that the caller to the fastrpc library does not know where the required files are located, such that the fastrpc library must search multiple places for them?

quic-parihar:
> Sorry, I think I've expressed this somewhere before but perhaps not in this PR.
> 
> Why do we need a search path at all? I understand that paths are different depending on the build target. But that can be resolved at build time. At runtime, why is it that the caller to the fastrpc library does not know where the required files are located, such that the fastrpc library must search multiple places for them?

I’m not entirely sure how this could be standardized across different distributions. Will it be right to control these at distro level?

@lumag to help add points to this on what should be done

basak-qcom:
> I’m not entirely sure how this could be standardized across different distributions. Will it be right to control these at distro level?

Yes, via a configure script option or environment variable or similar. The details would need to be worked out carefully between the library and the calling app, since I imagine that we don't want the library to access just any path on demand for security reasons.

Docs/conf_guideline.md
quic-parihar: acked

.github/workflows/build_linux_gnu.yml
```diff
@@ -20,11 +20,16 @@ jobs:
...
+          build-essential \
+          autoconf automake libtool pkg-config \
+          libyaml-dev \
+          tar gzip
 
     - name: Compile code for LE platform
       run: |
         GITCOMPILE_NO_MAKE=true ./gitcompile
-        make distcheck
+        make distcheck
```
ekanshibu: and empty line

inc/fastrpc_config_parser.h
```diff
@@ -0,0 +1,14 @@
...
+#define FASTRPC_YAML_PARSER_H
+
+#define DEFAULT_DSP_SEARCH_PATHS ";/usr/lib/rfsa/adsp;/vendor/lib/rfsa/adsp;/vendor/dsp/;/usr/lib/dsp/;"
+#define DSP_LIB_KEY "DSP_LIBRARY_PATH"
+
+extern char DSP_LIBS_LOCATION[PATH_MAX];
+
+void configure_dsp_paths();
+
+#endif /*FASTRPC_YAML_PARSER_H*/
```
ekanshibu: add empty line

src/fastrpc_config_parser.c
```diff
@@ -0,0 +1,177 @@
...
+  if (file == NULL)
+    return;
+
+  if (fgets(machine_name, sizeof(machine_name), file) != NULL)
+    // Remove trailing newline if present
+    machine_name[strcspn(machine_name, "\n")] = '\0';
+
+  fclose(file);
+  parse_config_dir(machine_name);
+}
```
ekanshibu: add empty line

quic-parihar: ack

inc/fastrpc_config_parser.h
quic-parihar: ack

.github/workflows/build_linux_gnu.yml
quic-parihar: ack

quic-parihar:
The changes have been made as discussed and marked complete. @lumag  @basak-qcom please have a look.

ekanshibu (dismissed):

Chennak-quic (dismissed):

src/Makefile.am
```diff
@@ -1,6 +1,12 @@
 lib_LTLIBRARIES =
 
-LIBDSPRPC_CFLAGS = -fno-short-enums -U_DEBUG -DARM_ARCH_7A -DLE_ENABLE -DENABLE_UPSTREAM_DRIVER_INTERFACE -DUSE_SYSLOG -I$(top_srcdir)/inc
+LIBDSPRPC_CFLAGS = -fno-short-enums -U_DEBUG -DARM_ARCH_7A -DLE_ENABLE -DENABLE_UPSTREAM_DRIVER_INTERFACE -DUSE_SYSLOG -DCONFIG_DIR='"/usr/share/qcom/conf.d/"' -DRELATIVE_LIB_PATH='"/usr/share/qcom/"' -DMACHINE_NAME_PATH='"/sys/firmware/devicetree/base/model"' -I$(top_srcdir)/inc
```
lumag: We have pointed out that `/usr/share/qcom/` prefix should be configurable. Also, there is no need for a separate `CONFIG_DIR`, use `RELATIVE_LIB_PATH "/conf.d/"`

quic-parihar: Acked

configure.ac
```diff
@@ -43,6 +43,26 @@ AM_PROG_CC_C_O
...
+])
+
+# Configure QCOM base path option (--with-qcom-path)
+AC_ARG_WITH([qcom-path],
+  [AS_HELP_STRING([--with-qcom-path=PATH],
+                  [Base directory for QCOM files (default: /usr/share/qcom)])],
+  [qcom_path="$withval"],
+  [qcom_path="/usr/share/qcom"])
+AC_MSG_NOTICE([QCOM path: $qcom_path])
+AC_SUBST([QCOM_PATH], ["$qcom_path"])
```
Chennak-quic: Could you rename QCOM_PATH to something more descriptive, such as CONFIG_BASE_DIR ?

quic-parihar: acked

Chennak-quic (dismissed):

inc/apps_std_internal.h
```diff
@@ -22,38 +22,19 @@
...
-#ifndef DSP_MOUNT_LOCATION
-#define DSP_MOUNT_LOCATION "/usr/lib/dsp/"
-#endif
-#ifndef DSP_DOM_LOCATION
-#define DSP_DOM_LOCATION "/usr/lib/dsp/xdspn"
-#endif
-#endif /* ENABLE_UPSTREAM_DRIVER_INTERFACE */
-
 #ifndef VENDOR_DSP_LOCATION
 #define VENDOR_DSP_LOCATION "/vendor/dsp/"
```
lumag, line 26: Drop all `/vendor` paths. They should never be used on Linux system.

inc/fastrpc_config_parser.h
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: BSD-3-Clause-Clear
+
+#ifndef FASTRPC_YAML_PARSER_H
+#define FASTRPC_YAML_PARSER_H
+
+#define DEFAULT_DSP_SEARCH_PATHS ";/usr/lib/rfsa/adsp;/vendor/lib/rfsa/adsp;/vendor/dsp/;/usr/lib/dsp/;"
```
lumag: None of these paths should be used, drop.

src/apps_std_imp.c
```diff
@@ -961,8 +965,10 @@ static int get_dirlist_from_env(const char *envvarname, char **ppDirList) {
         VERIFY(0 == (nErr = apps_std_getenv(envvarname, envList,
                                             envListPrependLen, &listLen)));
       }
+      // Append semicolon before DSP_SEARCH_PATH
+      strlcat(envList, ";", envListPrependLen);
       // Append default DSP_SEARCH_PATH to user defined env
-      strlcat(envList, DSP_SEARCH_PATH, envListPrependLen);
+      strlcat(envList, dsp_search_path, envListPrependLen);
```
lumag, line 971: So... First we concatenate the strings here, then you parse that during file open. Why? Can't you perform it exactly in the opposite way? Build the _list_ of paths here only **once**, without additional parsing of the "resulting" string?

src/fastrpc_apps_user.c
```diff
@@ -75,17 +75,13 @@
...
+#include "fastrpc_config_parser.h"
 
-#ifndef ENABLE_UPSTREAM_DRIVER_INTERFACE
-#define DSP_MOUNT_LOCATION "/dsp/"
-#define DSP_DOM_LOCATION "/dsp/xdsp"
-#else
-#define DSP_MOUNT_LOCATION "/usr/lib/dsp/"
-#define DSP_DOM_LOCATION "/usr/lib/dsp/xdspn"
-#endif
 #define VENDOR_DSP_LOCATION "/vendor/dsp/"
```
lumag, line 80: "/vendor" again, it should go.

src/fastrpc_apps_user.c
```diff
@@ -3459,27 +3455,13 @@ static int open_shell(int domain_id, apps_std_FILE *fh, int unsigned_shell) {
...
-                (absName = (char *)realloc(absName, sizeof(char) * absNameLen)),
-            AEE_ENOMEMORY);
-    strlcpy(absName, DSP_MOUNT_LOCATION, absNameLen);
-    strlcat(absName, SUBSYSTEM_NAME[domain], absNameLen);
-    strlcat(absName, "/", absNameLen);
-    strlcat(absName, shell_absName, absNameLen);
-    nErr = apps_std_fopen(absName, "r", fh);
-  }
   if (nErr) {
     absNameLen = strlen(VENDOR_DSP_LOCATION) + shell_absNameLen + 1;
```
lumag, line 3466: Drop

inc/apps_std_internal.h
quic-parihar: being dropped as part of this PR: https://github.com/qualcomm/fastrpc/pull/154

src/fastrpc_apps_user.c
quic-parihar: being dropped as part of this PR: https://github.com/qualcomm/fastrpc/pull/154

src/fastrpc_apps_user.c
quic-parihar: being dropped as part of this PR: https://github.com/qualcomm/fastrpc/pull/154,

We cannot completely remove /vendor path support from project as it will be supported for android platform as well. As part of above PR we will plan to move vendor specific macros and code to android specific file.

inc/fastrpc_config_parser.h
quic-parihar: These are default search paths we have been using till now.
These paths have been exposed to all the fastrpc clients via hexagon sdk documentation. Removing the support for these paths breaks backward compatibility/client expectations.

src/fastrpc_apps_user.c
lumag: You have the function which (should) return a list of search dirs (after parsing the config file and reading environment). On Android that function should return a fixed set of strings - `/vendor/dsp`, `/usr/lib/rfsa`, etc.

inc/fastrpc_config_parser.h
quic-parihar: From these paths we can remove vendor specific path and bring them as part of PR 154 where clean-up is being done and a separate file will be introduced for android specific things.

/usr/* paths we cannot remove as some LE platforms are already using those paths.

src/apps_std_imp.c
quic-parihar: Can you help elaborate more on this. When you are saying parsing are you referring to place where we are going through these paths and searching one by one in domain specific and fallback path option? If yes, then true we can prepare all paths here itself and then try to open in one go.

lumag: No, I'm saying that you are going through the string, looking for `;` symbols, etc. That needs to happen only once, while you handle the environment variable. After that it should be a list of strings. The path that you got from the config file should be added to the list rather than being concatenated to the string.

inc/fastrpc_config_parser.h
lumag: Nope, drop them here and bring that as an LE-specific patch so that those locations can die.

src/fastrpc_apps_user.c
quic-parihar: got it. Will include this in the same PR 154, where we are dropping this.

inc/fastrpc_config_parser.h
quic-parihar: Acked

src/apps_std_imp.c
quic-parihar: Due to urgency of this PR availability by other teams I will take this issue here: https://github.com/qualcomm/fastrpc/issues/270, as this is an issue present currently and not related to the change. 

Will start work on this starting next week to avoid any delays.

lumag: Ack, thanks.

lumag (approved):

PR #240, opened 2025-09-11
dsprpcd: dlopen versioned listener libraries with fallback

### LINKED ISSUE #230: dlopen() on X_default_listener.so is called without ABI versions
`*dsprpcd` binaries currently load a versionless `*_default_listener.so` (now from `src/dsprpcd.c`), but these have ABI versions defined, eg. `libadsp_default_listener.so.1`. Since the subsequent use of the `dlopen()`'d library assumes an ABI, the corresponding `dlopen()` call should explicitly specify the ABI version it requires by loading it by the name that includes the ABI version. This breaks expectations around shared library versioning in the global ecosystem namespace in `/usr/lib`. Further, we won't be able to concurrently ship two versions of the library following a future ABI version bump.

This affects tooling and the upcoming Debian builds, since Debian does not ship unversioned `.so` files in its runtime (non development) library packaging.

I can distro patch this for now, changing the `#define` to include a `.1` suffix. You could do that upstream, or if you want a more elaborate system to automatically build against the current ABI version, you could do that as an alternative. I suggest that ABI version bumps are rare enough that the latter isn't worth the maintenance burden, so the former is the best path forward. You will need also need to be confident that all builds are shipping the `.so.1` files. This seems likely, but I'll leave it to you to decide. If you're concerned, I guess you could implement a `dlopen()` fallback to try the unversioned name.

thiru31:
Fixes #230 
- Currently `dsprpcd.c` attempts to `dlopen()` unversioned libraries such as `libadsp_default_listener.so`,  This breaks on distributions  which only ship ABI-versioned libraries in runtime packages. 
- It also risks binding to a different ABI in the future if a .so.2 is introduced, as mentioned in #230.

This patch updates the default listener library names to include `.so.1` and adds a fallback to the un-versioned `.so` for environments that still ship only the un-versioned file.

### DIFF
--- a/src/dsprpcd.c
+++ b/src/dsprpcd.c
@@ -15,39 +15,80 @@
 #include <unistd.h>
 #include <string.h>
 
-#ifndef ADSP_DEFAULT_LISTENER_NAME
-#define ADSP_DEFAULT_LISTENER_NAME "libadsp_default_listener.so"
+#ifndef ADSP_LISTENER_VERSIONED
+#define ADSP_LISTENER_VERSIONED   "libadsp_default_listener.so.1"
+#define ADSP_LISTENER_UNVERSIONED "libadsp_default_listener.so"
 #endif
-#ifndef CDSP_DEFAULT_LISTENER_NAME
-#define CDSP_DEFAULT_LISTENER_NAME "libcdsp_default_listener.so"
+#ifndef CDSP_LISTENER_VERSIONED
+#define CDSP_LISTENER_VERSIONED   "libcdsp_default_listener.so.1"
+#define CDSP_LISTENER_UNVERSIONED "libcdsp_default_listener.so"
 #endif
-#ifndef SDSP_DEFAULT_LISTENER_NAME
-#define SDSP_DEFAULT_LISTENER_NAME "libsdsp_default_listener.so"
+#ifndef SDSP_LISTENER_VERSIONED
+#define SDSP_LISTENER_VERSIONED   "libsdsp_default_listener.so.1"
+#define SDSP_LISTENER_UNVERSIONED "libsdsp_default_listener.so"
 #endif
-#ifndef GDSP_DEFAULT_LISTENER_NAME
-#define GDSP_DEFAULT_LISTENER_NAME "libcdsp_default_listener.so.1"
+#ifndef GDSP_LISTENER_VERSIONED
+#define GDSP_LISTENER_VERSIONED   "libcdsp_default_listener.so.1"
+#define GDSP_LISTENER_UNVERSIONED "libcdsp_default_listener.so"
 #endif
 
 typedef int (*dsp_default_listener_start_t)(int argc, char *argv[]);
 
+// Result struct for dlopen.
+struct dlopen_result {
+  void *handle;
+  const char *loaded_lib_name;
+};
+
+/**
+ * Attempts to load a shared library using dlopen.
+ * If the versioned name fails, falls back to the unversioned name.
+ * Returns both the handle and the name of the library successfully loaded.
+ */
+static struct dlopen_result try_dlopen(const char *versioned, const char *unversioned) {
+  struct dlopen_result result = { NULL, NULL };
+
+  result.handle = dlopen(versioned, RTLD_NOW);
+  if (result.handle) {
+    result.loaded_lib_name = versioned;
+    return result;
+  }
+
+  if (unversioned) {
+    VERIFY_IPRINTF("dlopen failed for %s: %s; attempting fallback %s",
+                   versioned, dlerror(), unversioned);
+    result.handle = dlopen(unversioned, RTLD_NOW);
+    if (result.handle) {
+      result.loaded_lib_name = unversioned;
+      return result;
+    }
+  }
+  return result;
+}
+
 int main(int argc, char *argv[]) {
   int nErr = 0;
-  void *dsphandler = NULL;
-  const char* lib_name;
+  struct dlopen_result dlres = { NULL, NULL };
+  const char* lib_versioned;
+  const char* lib_unversioned;
   const char* dsp_name;
   dsp_default_listener_start_t listener_start;
 
   #ifdef USE_ADSP
-    lib_name = ADSP_DEFAULT_LISTENER_NAME;
+    lib_versioned = ADSP_LISTENER_VERSIONED;
+    lib_unversioned = ADSP_LISTENER_UNVERSIONED;
     dsp_name = "ADSP";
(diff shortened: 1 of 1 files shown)


src/dsprpcd.c
```diff
@@ -16,20 +16,34 @@
...
+#define CDSP_DEFAULT_LISTENER_FALLBACK "libcdsp_default_listener.so"
 #endif
 #ifndef SDSP_DEFAULT_LISTENER_NAME
-#define SDSP_DEFAULT_LISTENER_NAME "libsdsp_default_listener.so"
+#define SDSP_DEFAULT_LISTENER_NAME "libsdsp_default_listener.so.1"
+#define SDSP_DEFAULT_LISTENER_FALLBACK "libsdsp_default_listener.so"
 #endif
 #ifndef GDSP_DEFAULT_LISTENER_NAME
-#define GDSP_DEFAULT_LISTENER_NAME "libcdsp_default_listener.so.1"
+#define GDSP_DEFAULT_LISTENER_NAME "libgdsp_default_listener.so.1"
```
ekanshibu: there is no libgdsp* built as part of this project, please use libcdsp* for GDSP also.

src/dsprpcd.c
```diff
@@ -16,20 +16,34 @@
...
+#define GDSP_DEFAULT_LISTENER_FALLBACK "libgdsp_default_listener.so"
 #endif
 
 typedef int (*dsp_default_listener_start_t)(int argc, char *argv[]);
 
+static void *try_dlopen(const char *lib_name, const char *fallback) {
```
ekanshibu: add a function header, something like:
/**
 * Attempts to load a shared library using dlopen.
 * If the versioned name fails, falls back to the unversioned name.
 */

src/dsprpcd.c
```diff
@@ -16,20 +16,34 @@
...
+  void *handle = dlopen(lib_name, RTLD_NOW);
+  if (!handle && fallback) {
+    VERIFY_EPRINTF("Failed to load %s: %s", lib_name, dlerror());
+    VERIFY_EPRINTF("Trying fallback %s", fallback);
```
ekanshibu: VERIFY_EPRINTF --> VERIFY_IPRINTF to avoid log flooding

src/dsprpcd.c
```diff
@@ -55,7 +69,8 @@ int main(int argc, char *argv[]) {
   VERIFY_EPRINTF("%s daemon starting", dsp_name);
   
   while (1) {
-        if (NULL != (dsphandler = dlopen(lib_name,RTLD_NOW))) {
+        dsphandler = try_dlopen(lib_name, lib_fallback);
```
ekanshibu: where is lib_fallback defined?

src/dsprpcd.c
```diff
@@ -16,20 +16,34 @@
 #include <string.h>
 
 #ifndef ADSP_DEFAULT_LISTENER_NAME
-#define ADSP_DEFAULT_LISTENER_NAME "libadsp_default_listener.so"
+#define ADSP_DEFAULT_LISTENER_NAME "libadsp_default_listener.so.1"
+#define ADSP_DEFAULT_LISTENER_FALLBACK "libadsp_default_listener.so"
```
quic-parihar: Change the macro names to represent versioned and Un versioned libraries. Keep the macro names according to what they represent and not on code logic will help with readability and maintainability.

src/dsprpcd.c
quic-parihar: do we need multiple VERIFY print statements, try to keep them as one

src/dsprpcd.c
quic-parihar: why do we need a new lib_fallback variable? Shouldn't it be okay to fallback every time versioned so's are not present

src/dsprpcd.c
thiru31: Thanks for pointing out that, Fixed 👍

src/dsprpcd.c
thiru31: Added  relevant function header, please let me know if you like to have some changes

src/dsprpcd.c
thiru31: Fixed. Thanks for the suggestion 👍

src/dsprpcd.c
thiru31: Fixed.

src/dsprpcd.c
thiru31: Thanks for the suggestion @quic-parihar . This really helps in readability 💯

src/dsprpcd.c
thiru31: clubbed into single print statement.

src/dsprpcd.c
thiru31: IMO it’s useful to keep this for easy debugging, it will rarely fall back. Having it makes the codebase more self-documenting and gives better visibility if a fallback occurs.

However, I’m okay with whichever approach folks decide on.

@ekanshibu I would also like to get your inputs on this. thanks in advance

src/dsprpcd.c
```diff
@@ -15,54 +15,82 @@
...
+        dsphandler = try_dlopen(lib_versioned, lib_unversioned);
+        if (NULL != dsphandler) {
             if (NULL != (listener_start = (dsp_default_listener_start_t)dlsym(
                               dsphandler, "adsp_default_listener_start"))) {
                 VERIFY_IPRINTF("adsp_default_listener_start called");
                 nErr = listener_start(argc, argv);
             }
             if (0 != dlclose(dsphandler)) {
-              VERIFY_EPRINTF("dlclose failed for %s", lib_name);
+              VERIFY_EPRINTF("dlclose failed for %s", lib_versioned);
```
thiru31: I noticed that `dlclose()` currently always logs the versioned library name, even if the daemon actually loaded the `unversioned` fallback. That could be misleading during debugging.

One idea I have is to extend the helper so that it not only returns the `dlopen()` handle but also provides the name of the library that was successfully loaded. This way, we could log the correct name in `dlclose()` and in other messages.

Would this approach be acceptable, or do you have a different preference? I’d like to hear your POV. @ekanshibu @quic-parihar

quic-parihar: Suggestion: Improve dlopen helper to return actual loaded library name

try_dlopen function can return structure something like this

struct dlopenresult {
    void* handle;
    const char* loaded_lib_name;
};

thiru31: @quic-parihar, Done 👍.. can you help review this PR.

Chennak-quic (approved):

PR #243, opened 2025-09-16
Enable PD exception logging by default

quic-parihar:
Previously, PD exception logging was only enabled when the PD_EXCEPTION_LOGGING compile-time flag was defined. This change enables PD exception logging by default, allowing DSP fatal logs to be redirected to the shared buffer for apps-side logging.


src/fastrpc_apps_user.c
```diff
@@ -4040,15 +4040,13 @@ static int domain_init(int domain, int *dev) {
                 ret == (int)(DSP_AEE_EOFFSET + AEE_EUNSUPPORTED),
             ret);
```
ekanshibu, line 4041: write commit text in imperative style:
https://www.kernel.org/doc/html/latest/process/submitting-patches.html#describe-your-changes

quic-parihar: Updated commit text

ekanshibu (approved):

PR #249, opened 2025-10-15
Add GitHub Actions workflow for ABI/API compatibility check

quic-vkatoch:
Implement automated ABI/API compatibility checks on pull requests targeting development and main branches. Build AArch64 libraries using GNU cross-compilation toolchain, generate ABI dumps with abi-dumper for libadsprpc/libcdsprpc/libsdsprpc, and compare against baseline using abi-compliance-checker. Fail workflow on incompatibilities and upload detailed HTML reports as artifacts.

### DIFF
--- a/.github/workflows/abi-compat.yml
+++ b/.github/workflows/abi-compat.yml
@@ -1,0 +1,120 @@
+# =============================================================================
+# ABI/API Compatibility Check Workflow
+# =============================================================================
+# Ensures that PR changes don't break binary compatibility for existing apps
+# by comparing shared library ABIs between PR and development branch versions
+name: ABI/API Compatibility Check
+
+on:
+  pull_request:
+    branches: [development, main]
+
+# Cross-compilation environment for AArch64 (ARM64) architecture
+env:
+  CC: aarch64-linux-gnu-gcc
+  CXX: aarch64-linux-gnu-g++
+  AS: aarch64-linux-gnu-as
+  LD: aarch64-linux-gnu-ld
+  RANLIB: aarch64-linux-gnu-ranlib
+  STRIP: aarch64-linux-gnu-strip
+  
+  # -fno-eliminate-unused-debug-types prevents GCC from optimizing away debug info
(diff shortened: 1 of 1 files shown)


.github/workflows/abi-compat.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Run ABI compatibility check
+        shell: bash
+        run: |
+          mkdir -p abi/{dumps,reports,headers-{baseline,pr}}
+          
+          # Copy headers for ABI comparison
+          cp ../baseline/inc/{remote,rpcmem}.h abi/headers-baseline/
+          cp inc/{remote,rpcmem}.h abi/headers-pr/
+          
+          LIBS=(libadsprpc.so libcdsprpc.so libsdsprpc.so)
```
quic-mtharu, line 81: Do we need to review all three libraries, or is running it on one library sufficient?

.github/workflows/abi-compat.yml
```diff
@@ -0,0 +1,84 @@
...
+          FAIL=0
+          
+          for lib in "${LIBS[@]}"; do
+            base_lib="../baseline/src/.libs/$lib"
+            pr_lib="src/.libs/$lib"
+            name="${lib%%.so*}"
```
quic-mtharu, line 89: Please add comments to improve readability - `extract the base name of a shared library file, excluding its .so* version suffix`

.github/workflows/abi-compat.yml
quic-vkatoch: For now, I'm keeping all three libraries to avoid confusion for users. Reviewing just one might lead to ambiguity about which library to use, so including all ensures clarity at the user end. We can revisit this if a single library becomes the standard in future.

.github/workflows/abi-compat.yml
quic-vkatoch: Ack, I'll update.

Chennak-quic (approved):

PR #251, opened 2025-10-19
Add workflow to trigger a build of the package repo

simonbeaudoin0935:
This workflow is an adapted copy of this workflow : https://github.com/qualcomm-linux/pkg-template/blob/main/.github/workflows/to_paste_in_upstream/pkg-build-pr-check.yml

With this workflow in position, a PR in the upstream project repo (this one) triggers a build or the sister package repo (in this case, it will be https://github.com/qualcomm-linux/pkg-fastrpc). The special repo variable PKG_REPO_GITHUB_NAME will need to be set to "qualcomm-linux/pkg-fastrpc" in this repo for this to work. 

See visit https://github.com/qualcomm-linux/pkg-template for more info about this whole CI workflow.

When a PR is open in this repo, this workflow will in turn call this one :
https://github.com/qualcomm-linux/qcom-build-utils/blob/main/.github/workflows/qcom-upstream-pr-pkg-build-reusable-workflow.yml
What it will do is try to compile the debian package from the "qualcomm-linux/pkg-fastrpc" but with the changes from the PR opened on this side added in order to test that the debian package will still build if this PR were to be merged. 

Ultimately, there will be even more tests downstream from that, like integration tests in a whole distro.

### DIFF
--- a/.github/workflows/pkg-build-pr-check.yml
+++ b/.github/workflows/pkg-build-pr-check.yml
@@ -1,0 +1,39 @@
+
+name: Package Build PR Check
+description: |
+  This workflow will execute when a PR is open against the configured branch.
+  On top of whatever kind of CI/CD logic happens in this upstream repo, this 
+  workflow will make sure that what is attempted to be merged in the main release
+  branch won't break the debian package. The variable PKG_REPO_GITHUB_NAME needs to
+  be set in the repo where this file reside. This variable hold the packaging repo 
+  name on github associated to this upstream repo. What will happen is that the PR
+  triggering this workflow will go knock on the packaging repo's door and trigger
+  a full build of the package if it were to include these changes.
+
+on:
+  pull_request_target:
+    branches: [ main, development ]
+    types:
+      - ready_for_review
+      - synchronize
+    paths-ignore:
+      - '.github/**'
+
+permissions:
+ contents: read
+
+concurrency:
+  group: pr-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  package-build-pr-check:
+    uses: qualcomm-linux/qcom-build-utils/.github/workflows/qcom-upstream-pr-pkg-build-reusable-workflow.yml@main
+    with:
+      qcom-build-utils-ref: main
+      upstream-repo: ${{github.repository}}
+      upstream-repo-ref: ${{github.head_ref}}
+      pkg-repo: ${{vars.PKG_REPO_GITHUB_NAME}}
+      pr-number: ${{github.event.pull_request.number}}
+    secrets:
+      TOKEN: ${{ secrets.DEB_PKG_BOT_CI_TOKEN }}


.github/workflows/pkg-build-pr-check.yml
```diff
@@ -0,0 +1,31 @@
...
+  triggering this workflow will go knock on the packaging repo's door and trigger
+  a full build of the package if it were to include these changes.
+
+on:
+  pull_request_target:
+    branches: [ main, development ]
+
+permissions:
+ contents: read
+ security-events: write
```
ekanshibu: why security-events need to have write permissions? Is there any code scanning happening with this workflow?

.github/workflows/pkg-build-pr-check.yml
```diff
@@ -0,0 +1,31 @@
...
+
```
ekanshibu, line 28: Is concurrency check also needed here? Say if multiple PR updates happen quickly?

.github/workflows/pkg-build-pr-check.yml
```diff
@@ -0,0 +1,31 @@
...
+  On top of whatever kind of CI/CD logic happens in this upstream repo, this 
+  workflow will make sure that what is attempted to be merged in the main release
+  branch won't break the debian package. The variable PKG_REPO_GITHUB_NAME needs to
+  be set in the repo where this file reside. This variable hold the packaging repo 
+  name on github associated to this upstream repo. What will happen is that the PR
+  triggering this workflow will go knock on the packaging repo's door and trigger
+  a full build of the package if it were to include these changes.
+
+on:
+  pull_request_target:
```
ekanshibu, line 14: should it be `pull_request` instead of `pull_request_target`? Or is it intended to run the workflow against the base repo instead of the PR's repo? I'm assuming this workflow is intended for all the PRs that are getting raised for this project.

.github/workflows/pkg-build-pr-check.yml
```diff
@@ -0,0 +1,31 @@
...
+    branches: [ main, development ]
+
+permissions:
+ contents: read
+ security-events: write
+
+jobs:
+  package-build-pr-check:
```
ekanshibu, line 30: adding `timeout-minutes` here might help to avoid stuck runners

.github/workflows/pkg-build-pr-check.yml
simonbeaudoin0935: i think youre right, this was copied over from a template, I don't think we need here at all

.github/workflows/pkg-build-pr-check.yml
simonbeaudoin0935: hmm, I havent thought about this. I dont see why not, I'll add a concurrency statement

.github/workflows/pkg-build-pr-check.yml
simonbeaudoin0935: I have a habit of using 'pull_request_target' for security reason, so running the PR with the workflow from the basebranch, but using the content of the PR branch with the "upstream-repo-ref: ${{github.head_ref}}". What this does is that in the qcom-upstream-pr-pkg-build-reusable-workflow.yml reusable withflow, I checkout the content of the PR

.github/workflows/pkg-build-pr-check.yml
simonbeaudoin0935: hmm, thats not a bad idea, but it seems like its not possible to add it here when using a reusable workflow, it would have to be in a job with the steps: statement, aka inside the reusable workflow

ekanshibu (approved):

.github/workflows/pkg-build-pr-check.yml
```diff
@@ -0,0 +1,34 @@
...
+  workflow will make sure that what is attempted to be merged in the main release
+  branch won't break the debian package. The variable PKG_REPO_GITHUB_NAME needs to
+  be set in the repo where this file reside. This variable hold the packaging repo 
+  name on github associated to this upstream repo. What will happen is that the PR
+  triggering this workflow will go knock on the packaging repo's door and trigger
+  a full build of the package if it were to include these changes.
+
+on:
+  pull_request_target:
+    branches: [ main, development ]
```
quic-mtharu, line 15: We should ignore changes under .github/** to avoid unnecessary CI runs triggered by workflow‑only updates.
Please add a paths-ignore: entry so CI doesn’t run when only workflow files are modified.

.github/workflows/pkg-build-pr-check.yml
quic-mtharu: **Can we add types: [opened, synchronize, reopened] to this workflow?**

Without specifying these event types, the workflow will trigger on every PR event — including label changes, description edits, reviewer assignments, etc. This leads to unnecessary CI runs and wasted build minutes.

Adding the types filter ensures the workflow only runs when it actually matters (new PR, new commits, or a reopened PR), keeping CI faster and more efficient.

PR #253, opened 2025-11-03
inc: add required include files at instal

Srinivas-Kandagatla:
Install  AEEStdDef.h, HAP_debug.h  and  HAP_farf.h that are required by to compile the skel and stub files generated by qaic compiler.

qaic compiler includes these header files, however fastrpc does not install these files, fix this by installing the required files.

### DIFF
--- a/inc/Makefile.am
+++ b/inc/Makefile.am
@@ -1,16 +1,16 @@
 # Export fastrpc headers
 fastrpc_includedir       = $(includedir)/fastrpc
 fastrpc_include_HEADERS  = $(top_srcdir)/inc/AEEStdErr.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/AEEStdDef.h
 fastrpc_include_HEADERS += $(top_srcdir)/inc/remote.h
 fastrpc_include_HEADERS += $(top_srcdir)/inc/rpcmem.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/HAP_farf.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/HAP_debug.h
 
 noinst_HEADERS = \
 	AEEBufBound.h \
 	AEEQList.h \
-	AEEStdDef.h \
 	AEEstd.h \
-	HAP_debug.h \
-	HAP_farf.h \
 	HAP_farf_internal.h \
 	HAP_pls.h \
 	adsp_current_process.h \


inc/Makefile.am
```diff
@@ -1,16 +1,16 @@
 # Export fastrpc headers
 fastrpc_includedir       = $(includedir)/fastrpc
 fastrpc_include_HEADERS  = $(top_srcdir)/inc/AEEStdErr.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/AEEStdDef.h
 fastrpc_include_HEADERS += $(top_srcdir)/inc/remote.h
 fastrpc_include_HEADERS += $(top_srcdir)/inc/rpcmem.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/HAP_farf.h
+fastrpc_include_HEADERS += $(top_srcdir)/inc/HAP_debug.h
```
lumag, line 8: I'm fine with the first one. Do the last two make the actual API of the library or should we kill them from the qaic output?

Srinivas-Kandagatla: if the modules are adding some debug prints into it then these headers are required.

lumag: Why are they a part of the FastRPC library then? In the end, the FastRPC is used for building the _host_ code, not the DSP code.

Srinivas-Kandagatla: Yes, that is one of the point that I raised before too, some of the headerfiles in the fastrpc are  DSP header files, these do not belong to fastrpc. I initial suggestion was to release them as dsp dev headers which should enable any library  (even fastrpc) or application  to use these headers. Lets work on cleaning this up.

ekanshibu: > Yes, that is one of the point that I raised before too, some of the headerfiles in the fastrpc are DSP header files, these do not belong to fastrpc. I initial suggestion was to release them as dsp dev headers which should enable any library (even fastrpc) or application to use these headers. Lets work on cleaning this up.

Just to classify different types of headers:
1. Core fastrpc library headers (fastrpc_ioctl.h, fastrpc_apps_user.h etc.)
2. Common headers (Used by application: HAP_*.h, verify.h etc.)

Where do you suggest we place the headers that provides fastrpc APIs, like remote.h, rpcmem.h, dspqueue.h etc? Should it be kept in this project and just installed during make install?

Also for common header(2), you are suggesting a separate project which can be used by different users, right?

Srinivas-Kandagatla: remote.h, AEE*.h and HAP*.h are dsp header files
fatrpc_ioctl.h this should never be part of fastrpc library, this comes from kernel-headers package.
rpcmem.h is fastrpc header it should be installed by fastrpc

ekanshibu: How are you differentiating b/w remote.h and rpcmem.h here? Both are available for applications. Both needs to maintain backward compatibility. So in terms of maintenance, both should come from same place as per my understanding.

Srinivas-Kandagatla: Correct me if am wrong here, but understanding is that rpcmem.h is the interface for implementation for allocating rpc mem. It can be implemented anywhere in fastrpc or as standalone lib. This has no interaction with DSP I assume.
BTW is rpcmem.h sourced from dsp builds? 

remote.h on the other hand is dsp header,.

ekanshibu: > Correct me if am wrong here, but understanding is that rpcmem.h is the interface for implementation for allocating rpc mem. It can be implemented anywhere in fastrpc or as standalone lib. This has no interaction with DSP I assume. 

Yes, rpcmem.h is providing APIs to allocate(and register) memory. Didn't exactly get what do you mean when you say that it can be implemented anywhere, do you mean that I can build it standalone without fastrpc lib, then I don't think it would work as every rpcmem allocated buffer has to be registered with fastrpc framework so that it is not treated as copy buffer by driver.

> BTW is rpcmem.h sourced from dsp builds?

No, none of the headers are actually sourced from DSP build. AEE*.h, HAP*.h and verify.h are the ones which are common between library and DSP. But those are not sourced from DSP. remote.h and rpcmem.h are also present on DSP build but those are not same as the one present here(let me know if you need more details on this). So, I don't think it can be sourced from DSP.

> 
> remote.h on the other hand is dsp header,.

I'm assuming when you say DSP header, you mean headers sourced from DSP build, then no, remote.h is not a DSP header as the DSP side file only has some basic DSP specific functionalities like(remote_handle_open, remote_handle_invoke etc.). Please correct me if my assumption is wrong.

Srinivas-Kandagatla: From the looks of it, remote.h looks like a FastRPC ABI between DSP and CPU (in our case), Am surprised to hear that it is not part of the dsp build, as every bit in that header file looks like an ABI.

If they are not part of DSP then you are highly likely to be out of sync over time and break the compatibility. I still think that what ever you are adding apart from what DSP is providing in that header file should be seperate header file and keep whatever dsp provides as unmodified and consider that as ABI headers.  I think we should discuss on this bit more.

for rpcmem.h,  Am still not clear on what DSP header looks like, but the apis required  are already in remote.h to register buffers remote_register_buf/remote_register_buf_attr. Let me look at the dsp header and comment more on this.

When I mean DSP headers, I mean DSP ABI's.

ekanshibu: Can I go ahead and merge this change?

PR #256, opened 2025-11-05
Fix PD exception logging

Jianping-Li:
When PD exception logging is enabled by default, RootPD daemons will be blocked as adspmsgd_init will wait for listener which is not initialized by the time it is called. Initialize listener before adspmsgd intialization.


src/fastrpc_apps_user.c
```diff
@@ -4040,13 +4040,6 @@ static int domain_init(int domain, int *dev) {
                 ret == (int)(DSP_AEE_EOFFSET + AEE_EUNSUPPORTED),
```
ekanshibu, line 4040: Can you give some details about the failure in the commit message?

PR #257, opened 2025-11-05
Clean headers

lumag:
Following the discussion in #253 , reorg headers and source files a bit.


inc/misc/fastrpc.h
Srinivas-Kandagatla, line 1: why are we maintaining this header file here, it should be part of kernel headers installed in the system. Looks redundant to me and will be out of sync over time.

Srinivas-Kandagatla (approved):
lgtm

inc/misc/fastrpc.h
lumag: You can check the history of the pipelines and commits here :D

Some of the pipelines have old compiler with no <misc/fastrpc.h> header. When I tried importing the header here (as it is recommended generally), the repolinter complained about importing GPL header into the BSD repo.

lumag: Oh. I see why it's failing.

src/Makefile.am
```diff
@@ -98,10 +98,12 @@ libsdsp_default_listener_la_CFLAGS = $(SDSP_CFLAGS) -DUSE_SYSLOG
 
 GDSP_CFLAGS = $(LIBDSPRPC_CFLAGS) -DDEFAULT_DOMAIN_ID=5
```
ekanshibu, line 100: libgdsprpc was actually intended to not be added as we are sticking with multidomain-only approach for gdsp. Here domain is always passed explicitly and no need to rely on library to extract domain information. So we can basically reuse any existing library and just pass the domain name to make it work. (example: ./gdsprpcd rootpd gdsp0)

I think it would be better to remove libgdsprpc_la* and libgdsp_default_listener* completely and just build gdsprpcd.

lool: https://github.com/qualcomm/fastrpc/pull/330/changes/4186401e5690a13d330cc7b7bd6320416a3cbe6c

Issue #245, opened 2025-09-17
Add services and udev rules file to fastrpc project


ekanshibu:
Add services for adsprpcd, cdsprpcd, sdsprpcd(with sensors PD argument) and gdsprpcd. Also add udev rules file to start the service upon fastrpc device creation.

ekanshibu:
Till now the service files have been enabled by default on downstream with Restart=always, this is done to ensure that the daemons are started automatically at boot.

There services were getting installed based on MACHINE(ex: CDSP1 and GDSP1 services are only needed for LeMans).

Now as we don't want to keep any MACHINE specific installation, we thought of installing all the services and start it based on the fastrpc device availability by adding a SYSTEMD_WANTS in udev rules file.

I understand this might not be practiced in general, so would be open for any suggestions/discussions for this. Till now on meta-qcom, the systemd service auto-enable is marked as "disable", so is it not expected to have the services auto-enabled by default? If that is the case, then we can just have the service installed from fastrpc project and control the enablement from the distro.

lumag:
Use existing features provided by systemd unit files instead of reimplementing them via udev.

PR #263, opened 2025-12-08
Add systemd service files for FastRPC daemons

quic-vkatoch:
Create systemd service files for DSP RPC daemons (aDSP, cDSP, gDSP, sDSP) with device-conditional startup.

Each service uses After directives to ensure proper ordering with corresponding device units (dev-fastrpc-\*-device and dev-fastrpc-\*-secure.device),and ConditionPathExists directives to conditionally start only when at least one of the device paths exists (/dev/fastrpc-\* or /dev/fastrpc-\*-secure). Includes automatic restart on failure with 5-second delay between attempts. Services are enabled for multi-user.target.

Fixes: #245

### DIFF
--- a/.github/workflows/build_linux_gnu.yml
+++ b/.github/workflows/build_linux_gnu.yml
@@ -32,4 +32,4 @@ jobs:
     - name: Compile code for LE platform
       run: |
         GITCOMPILE_NO_MAKE=true ./gitcompile
-        make distcheck
+        make distcheck DISTCHECK_CONFIGURE_FLAGS="--with-systemdsystemunitdir='lib/systemd/system'"


configure.ac
```diff
@@ -39,6 +39,19 @@ AM_PROG_CC_C_O
...
+# Check for systemd and get systemd unit directory
+PKG_CHECK_MODULES([SYSTEMD], [systemd], [have_systemd=yes], [have_systemd=no])
+if test "x$have_systemd" = "xyes"; then
+    AC_MSG_CHECKING([for systemd system unit directory])
+    systemdsystemunitdir=$($PKG_CONFIG --variable=systemdsystemunitdir systemd)
+    AC_MSG_RESULT([$systemdsystemunitdir])
+else
+    # Fallback to default systemd directory if pkg-config is not available
+    systemdsystemunitdir=/lib/systemd/system
+    AC_MSG_WARN([systemd not found via pkg-config, using default directory: $systemdsystemunitdir])
```
lumag: No, let users pass it directly via config options.

files/adsprpcd.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=aDSP RPC daemon
+BindsTo=dev-fastrpc-adsp-secure.device
+After=dev-fastrpc-adsp-secure.device
```
lumag: Why is it -secure?

quic-vkatoch: Will update

configure.ac
quic-vkatoch: Ack

PR #272, opened 2025-12-23
Build domain-correct paths and clean up shell open logs

quic-parihar:
Use SUBSYSTEM_NAME[domain] instead of the hard-coded "adsp" when computing the absolute file name in fopen_from_dirlist(), so paths are constructed for the actual domain in use. This fixes incorrect lookups on multi domain scenarios.

In open_shell(), remove redundant ALWAYS logs.


ekanshibu:
configure: error: libyaml (yaml-0.1) is required but not found.
 
 Checks are failing with this, can you pls fix this as well?

PR #273, opened 2025-12-30
Configure to run armor on aws runner

ayukum-hyd:
Use aws runners in run_armor workflow, which is need to upload artifacts generated in RUN-ARMOR workflow to s3 location

### DIFF
--- a/.github/workflows/run_armor.yml
+++ b/.github/workflows/run_armor.yml
@@ -37,7 +37,9 @@ permissions:
 
 jobs:
   RUN-ARMOR:
-    runs-on: ubuntu-22.04
+    runs-on:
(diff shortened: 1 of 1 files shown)


.github/workflows/run_armor.yml
```diff
@@ -69,4 +71,4 @@ jobs:
           base-sha:   ${{ steps.ev.outputs.base_sha }}
           ref:        ${{ steps.ev.outputs.ref }}
           repo:       ${{ steps.ev.outputs.repo }}
-          branch-name: ${{ steps.ev.outputs.branch_name }}
\ No newline at end of file
+          branch-name: ${{ steps.ev.outputs.branch_name }}
```
quic-aditrath, line 74: Seems no change here.

PR #275, opened 2026-01-05
Add systemd unit file for audiopd aDSP RPC daemon

qti-sbojja:
Add a systemd service to manage the audiopd RPC daemon to support dynamic module loading on the aDSP.

### DIFF
--- a/files/Makefile.am
+++ b/files/Makefile.am
@@ -6,6 +6,7 @@ systemdsystemunitdir = @systemdsystemunitdir@
 # List of systemd service files to install
 dist_systemdsystemunit_DATA = \
 	adsprpcd.service \
+	adsprpcd_audiopd.service \
 	cdsprpcd.service \
 	cdsp1rpcd.service \
 	gdsp0rpcd.service \
--- a/files/adsprpcd_audiopd.service
+++ b/files/adsprpcd_audiopd.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=audiopd aDSP RPC daemon
+After=dev-fastrpc-adsp.device dev-fastrpc-adsp-secure.device
+ConditionPathExists=|/dev/fastrpc-adsp
+ConditionPathExists=|/dev/fastrpc-adsp-secure
+
+[Service]
+Type=exec
+ExecStart=/usr/bin/adsprpcd audiopd
+Restart=on-failure
+RestartSec=5
+
+[Install]
+WantedBy=multi-user.target


lumag (changes requested):
Why do we need a separate unit file for this? Can adsprpcd start it during a normal startup procedure?

ekanshibu:
> Why do we need a separate unit file for this? Can adsprpcd start it during a normal startup procedure?

We need 2 different/separate daemon processes for root PD and audio PD. Are you suggesting starting both the processes from the same unit file? If this is the case then I don't see any major problem with that. As per my understanding, till now the reason for having service files separated for each daemon process is to have some level of control in case any particular daemon is not required on any platform.

lumag:
> > > Why do we need a separate unit file for this? Can adsprpcd start it during a normal startup procedure?
> > 
> > 
> > We need 2 different/separate daemon processes for root PD and audio PD. Are you suggesting starting both the processes from the same unit file? If this is the case then I don't see any major problem with that. As per my understanding, till now the reason for having service files separated for each daemon process is to have some level of control in case any particular daemon is not required on any platform.
> 
> @lumag, can you please confirm if adsprpcd daemon need to start audio/sensor PD's, we can make related changes and close this PR.

After a second thought, your current approach is fine. Please add necessary details to the commit message.

qti-sbojja:
> > > > Why do we need a separate unit file for this? Can adsprpcd start it during a normal startup procedure?
> > > 
> > > 
> > > We need 2 different/separate daemon processes for root PD and audio PD. Are you suggesting starting both the processes from the same unit file? If this is the case then I don't see any major problem with that. As per my understanding, till now the reason for having service files separated for each daemon process is to have some level of control in case any particular daemon is not required on any platform.
> > 
> > 
> > @lumag, can you please confirm if adsprpcd daemon need to start audio/sensor PD's, we can make related changes and close this PR.
> 
> After a second thought, your current approach is fine. Please add necessary details to the commit message.

I’ve updated the commit message. Let me know if any changes are needed. Thanks for reviewing.

lumag:
> > > > > Why do we need a separate unit file for this? Can adsprpcd start it during a normal startup procedure?
> > > > 
> > > > 
> > > > We need 2 different/separate daemon processes for root PD and audio PD. Are you suggesting starting both the processes from the same unit file? If this is the case then I don't see any major problem with that. As per my understanding, till now the reason for having service files separated for each daemon process is to have some level of control in case any particular daemon is not required on any platform.
> > > 
> > > 
> > > @lumag, can you please confirm if adsprpcd daemon need to start audio/sensor PD's, we can make related changes and close this PR.
> > 
> > 
> > After a second thought, your current approach is fine. Please add necessary details to the commit message.
> 
> I’ve updated the commit message. Let me know if any changes are needed. Thanks for reviewing.

I can repeat my (our) usual guidelines:
- start your commit message by describing the problem
- continue by describing what needs to be done
- don't describe patch contents

ekanshibu (approved):

Issue #265, opened 2025-12-11
Compilation problems and not working binary under linux:  implicit declaration of function 'strlcpy'


laklimov:
When compiling under Linux there a lot of warnings like this:

```
...
log_config.c: In function 'parseLogConfig':
log_config.c:105:3: warning: implicit declaration of function 'strlcpy'; did you mean 'strncpy'? [-Wimplicit-function-declaration]
  105 |   strlcpy(tempFiles, filenames, strlen(filenames) + 1);
      |   ^~~~~~~
      |   strncpy
mv -f .deps/libadsprpc_la-log_config.Tpo .deps/libadsprpc_la-log_config.Plo
/bin/bash ../libtool  --tag=CC   --mode=compile aarch64-linux-gnu-gcc -DHAVE_CONFIG_H -I. -I..    -fno-short-enums -U_DEBUG -DARM_ARCH_7A -DLE_ENABLE -DENABLE_UPSTREAM_DRIVER_INTERFACE -DUSE_SYSLOG -I../inc -DDEFAULT_DOMAIN_ID=0 -O2 -Wall -pipe -g -MT libadsprpc_la-dspsignal.lo -MD -MP -MF .deps/libadsprpc_la-dspsignal.Tpo -c -o libadsprpc_la-dspsignal.lo `test -f 'dspsignal.c' || echo './'`dspsignal.c
libtool: compile:  aarch64-linux-gnu-gcc -DHAVE_CONFIG_H -I. -I.. -fno-short-enums -U_DEBUG -DARM_ARCH_7A -DLE_ENABLE -DENABLE_UPSTREAM_DRIVER_INTERFACE -DUSE_SYSLOG -I../inc -DDEFAULT_DOMAIN_ID=0 -O2 -Wall -pipe -g -MT libadsprpc_la-dspsignal.lo -MD -MP -MF .deps/libadsprpc_la-dspsignal.Tpo -c dspsignal.c  -fPIC -DPIC -o .libs/libadsprpc_la-dspsignal.o
mv -f .deps/libadsprpc_la-dspsignal.Tpo .deps/libadsprpc_la-dspsignal.Plo
[...]

PR #279, opened 2026-01-19
Add libbsd support for non-Android builds and update CI workflows

quic-vkatoch:
Check for libbsd on non-Android targets and export BSD_CFLAGS and BSD_LIBS. Include BSD flags in CFLAGS and link BSD libs alongside YAML libs.

Add libbsd-dev:arm64 to ARM64 workflows and ensure libbsd-dev is in build_linux_gnu.yml.

## Technical Details

@BSD_CFLAGS@ adds the BSD include path (e.g., `-isystem /usr/include/bsd`). With libbsd-overlay, when `<string.h>` is included, the BSD header chain is pulled in indirectly through the overlay mechanism.

As a result, the compiler sees the strlcpy/strlcat prototypes and no longer warns about implicit declarations, without requiring an explicit `#include <bsd/string.h>`.

## Changes

- **configure.ac**: Added PKG_CHECK_MODULES for libbsd-overlay on non-Android builds
- **src/Makefile.am**: Added @BSD_CFLAGS@ to LIBDSPRPC_CFLAGS and @BSD_LIBS@ to library link flags
- **Workflows**: Updated CI workflows to install libbsd-dev package for Linux builds

Fixes: #265

### DIFF
--- a/.github/workflows/abi-compat.yml
+++ b/.github/workflows/abi-compat.yml
@@ -82,7 +82,8 @@ jobs:
             automake autoconf libtool pkg-config \
             gcc-aarch64-linux-gnu g++-aarch64-linux-gnu binutils-aarch64-linux-gnu \
             libyaml-dev \
-            libyaml-0-2:arm64 libyaml-dev:arm64
+            libyaml-0-2:arm64 libyaml-dev:arm64 \
+            libbsd-dev:arm64
 
           # Sanity checks
           abi-compliance-checker -version
--- a/.github/workflows/build_linux_arm64.yml
+++ b/.github/workflows/build_linux_arm64.yml
@@ -63,7 +63,8 @@ jobs:
           automake autoconf libtool pkg-config \
           gcc-aarch64-linux-gnu g++-aarch64-linux-gnu binutils-aarch64-linux-gnu\
           libyaml-dev \
-          libyaml-0-2:arm64 libyaml-dev:arm64
+          libyaml-0-2:arm64 libyaml-dev:arm64 \
+          libbsd-dev:arm64
 
     - name: Sanity check libyaml (headers + ARM64 pc)
       run: |
--- a/.github/workflows/build_linux_gnu.yml
+++ b/.github/workflows/build_linux_gnu.yml
@@ -26,7 +26,7 @@ jobs:
         sudo apt-get install -y \
           build-essential \
           autoconf automake libtool pkg-config \
-          libyaml-dev \
+          libyaml-dev libbsd-dev \
           tar gzip
 
     - name: Compile code for LE platform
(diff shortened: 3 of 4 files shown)


.github/workflows/abi-compat.yml
```diff
@@ -82,7 +82,8 @@ jobs:
             automake autoconf libtool pkg-config \
             gcc-aarch64-linux-gnu g++-aarch64-linux-gnu binutils-aarch64-linux-gnu \
             libyaml-dev \
-            libyaml-0-2:arm64 libyaml-dev:arm64
+            libyaml-0-2:arm64 libyaml-dev:arm64 \
+            libbsd-dev:arm64
```
quic-mtharu, line 86: Please add libbsd-dev installation to the Dockerfile and update the documentation.

configure.ac
```diff
@@ -63,6 +63,14 @@ AS_IF([test "$compile_for_android" = no], [
   AC_SUBST(YAML_LIBS)
 ])
 
+# Check for libbsd only if not Android (provides strlcpy, strlcat, etc.)
```
quic-mtharu, line 66: Should we add #include <bsd/string.h> to the source?

Linking with -lbsd only provides the symbol, but without the header the compiler treats strlcpy() as an implicit declaration, which leads to warnings and an incorrect function signature on Linux.

quic-vkatoch: **@BSD_CFLAGS@** adds the BSD include path (e.g., **-isystem /usr/include/bsd**), and with <string.h> already included, the BSD header chain is pulled in indirectly. 
As a result, the compiler now sees the strlcpy/strlcat prototypes and no longer warns about implicit declarations, even without an explicit #include <bsd/string.h>.

.github/workflows/abi-compat.yml
quic-vkatoch: Sure

configure.ac
quic-mtharu: It would be better to include this explanation in the PR description.

src/Makefile.am
```diff
@@ -151,4 +151,4 @@ gdsprpcddir = $(libdir)
 gdsprpcd_SOURCES = dsprpcd.c
 gdsprpcd_DEPENDENCIES = libcdsp_default_listener.la
 gdsprpcd_CFLAGS = -I$(top_srcdir)/inc -DDEFAULT_DOMAIN_ID=5 -DUSE_SYSLOG -DUSE_GDSP
-gdsprpcd_LDADD =  -ldl $(USE_LOG)
\ No newline at end of file
+gdsprpcd_LDADD =  -ldl $(USE_LOG)
```
lumag: Unrelated

lumag (approved):

src/Makefile.am
quic-vkatoch: Updated

Issue #295, opened 2026-02-15
Mermaid for diagrams


lool:
### FastRPC version

tip

### DSP firmware version

n/a

### SoC and device name

n/a

### Kernel version

n/a

### What happened?

Hi

Thanks for adding nice architecture diagrams! It would be nice to have them as text, and GitHub supports Mermaid diagrams natively, would it make sense to switch to those?

Here's a tentative for the main one

```mermaid
 flowchart LR
    subgraph SharedDriver[Shared Memory Driver]
        direction LR
        subgraph CPU
            direction TB
            subgraph Application
                direction TB
                Stub --> FastRPCUser[FastRPC user driver]
                style Stub fill:#88c
                style FastRPCUser fill:#44f
            end
            style Application fill:#888
            subgraph Kernel
                direction TB
                FastRPCKernel[FastRPC kernel driver]
                style FastRPCKernel fill:#44f
            end
        end
        Application --> Kernel
        subgraph DSP
            direction BT
            subgraph dspdriver[ ]
                FastRPCDSP[FastRPC DSP driver]
                style FastRPCDSP fill:#44f
            end
            subgraph UserPD[User PD]
                Skel --> FastRPCDSPuser[FastRPC DSP user driver]
                style Skel fill:#88c
            end
            style UserPD fill:#888
            style FastRPCDSPuser fill:#44f
            dspdriver --> UserPD
        end
        CPU --> DSP
    end
```

### Steps to reproduce

N/A

### Relevant log output

```shell

```

PR #296, opened 2026-02-16
README: Use Markdown to define Terms

lool:
Drop image which breaks the overall style and makes reading harder.


lool:
I would have liked to submit the two diagrams as mermaid, but these will require some stylistic review as I don't think mermaid can provide exact replacements; see #295.

lool:
If you're happy with the diagram I proposed in #295, I can add it to this PR, and try to draw the last remaining one which is simpler

ekanshibu (approved):

Chennak-quic (approved):

PR #302, opened 2026-02-24
Remove DEFAULT_DEVICE fallback to prevent cross-domain routing

### LINKED ISSUE #164: Remove fallback to default device open
In case of default domain ID, the device open request falls back to DEFAUL_DEVICE which will be defined in the libs to which the application is linked.

Remove this legacy fallback logic to avoid any confusion and unexpected observations.

Scenario:

If any application is linked to libadsprpc.so and user passed request to offload to CDSP, for general multi-domain calls, the flow will work as expected. But for legacy non-domain calls, the call might end up using adsp device(as DEFAULT_DEVICE for libadsprpc is fastrpc-adsp). Avoid this path and rely on the domain requested by the user.

1 more scenario is the unavailability of the device nodes for any requested domain. In the above scenario, if cdsp device nodes were not available, there will still be a fallback to adsp device node.

quic-vkatoch:
Remove legacy DEFAULT_DEVICE fallback that could cause incorrect domain routing. Previously, if a requested domain's device nodes were unavailable, the code would fall back to the DEFAULT_DEVICE of the linked library (e.g., ADSP for libadsprpc.so), ignoring the user's domain request.

Now each domain only attempts: secure node -> non-secure node, and fails cleanly if both are unavailable. This ensures device opening respects the user's requested domain.

Consolidate all domain cases (ADSP/MDSP/SDSP/CDSP/GDSP) into unified logic and improve error handling.

Additionally, merge domain device name functions into a unified API by replacing separate get_secure_domain_name() and get_non_secure_domain_name() implementations with a single get_domain_device_names() function that returns both names in one call.

Supersedes #210 due to outdated and stale changes.
Fixes: #164

### DIFF
--- a/inc/fastrpc_internal.h
+++ b/inc/fastrpc_internal.h
@@ -16,6 +16,7 @@
 #include "AEEQList.h"
 #include "AEEStdErr.h"
 #include "fastrpc_common.h"
+#include "fastrpc_ioctl.h"
 
 // Aligns the memory
 #define ALIGN_B(p, a)	      (((p) + ((a) - 1)) & ~((a) - 1))
@@ -450,6 +451,14 @@ static __inline int convert_kernel_to_user_error(int nErr, int err_no) {
 }
 
 /**
+  * @brief Get both secure and non-secure device names for a domain
+  * @param domain_id: Domain ID
+  * @param secure_name: Output pointer for secure device name
+  * @param non_secure_name: Output pointer for non-secure device name
+  **/
+void get_domain_device_names(int domain_id, const char **secure_name, const char **non_secure_name);
+
+/**
   * @brief utility APIs used in fastRPC library to get name, handle from domain 
   **/
 int get_domain_from_name(const char *uri, uint32_t type);
@@ -514,9 +523,5 @@ int ioctl_optimization(int dev, uint32_t max_concurrency);
  */
 int ioctl_mdctx_manage(int dev, int req, void *user_ctx,
 	unsigned int *domain_ids, unsigned int num_domain_ids, uint64_t *ctx);
-
-const char* get_secure_domain_name(int domain_id);
-
-#include "fastrpc_ioctl.h"
 
 #endif // FASTRPC_INTERNAL_H
--- a/src/adsp_default_listener.c
+++ b/src/adsp_default_listener.c
@@ -33,20 +33,6 @@
 #define POLL_TIMEOUT	10 * 1000
 #define EVENT_SIZE		( sizeof (struct inotify_event) )
 #define EVENT_BUF_LEN	( 1024 * ( EVENT_SIZE + 16 ) )
-#define ADSP_SECURE_DEVICE_NAME "fastrpc-adsp-secure"
-#define SDSP_SECURE_DEVICE_NAME "fastrpc-sdsp-secure"
-#define MDSP_SECURE_DEVICE_NAME "fastrpc-mdsp-secure"
-#define CDSP_SECURE_DEVICE_NAME "fastrpc-cdsp-secure"
-#define CDSP1_SECURE_DEVICE_NAME "fastrpc-cdsp1-secure"
-#define GDSP0_SECURE_DEVICE_NAME "fastrpc-gdsp0-secure"
-#define GDSP1_SECURE_DEVICE_NAME "fastrpc-gdsp1-secure"
-#define ADSP_DEVICE_NAME "fastrpc-adsp"
-#define SDSP_DEVICE_NAME "fastrpc-sdsp"
-#define MDSP_DEVICE_NAME "fastrpc-mdsp"
-#define CDSP_DEVICE_NAME "fastrpc-cdsp"
-#define CDSP1_DEVICE_NAME "fastrpc-cdsp1"
-#define GDSP0_DEVICE_NAME "fastrpc-gdsp0"
-#define GDSP1_DEVICE_NAME "fastrpc-gdsp1"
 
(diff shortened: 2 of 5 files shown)


src/adsp_default_listener.c
```diff
@@ -97,15 +97,15 @@ static const char *get_secure_device_name(int domain_id) {
...
+		FARF(ERROR, "ERROR: %s Invalid domain %d", __func__, domain);
 		break;
 	}
 
 	return name;
 }
 
-static const char *get_default_device_name(int domain_id) {
-	const char *name;
+static const char *get_non_secure_device_name(int domain_id) {
```
quic-anane: Can we see if we can reduce the number of functions ? 
some are just returning same string with /dev/ removed , can we write a wrapper with string operations to do the same from one function ?

src/adsp_default_listener.c
```diff
@@ -172,13 +172,13 @@ static bool fastrpc_dev_exists(const char* dev_name)
 static int fastrpc_wait_for_device(int domain)
 {
 	int inotify_fd = -1, watch_fd = -1, err = 0;
-	const char *sec_dev_name = NULL, *def_dev_name = NULL;
+	const char *sec_dev_name = NULL, *non_sec_dev_name = NULL;
 	struct pollfd pfd[1];
 
 	sec_dev_name = get_secure_device_name(domain);
-	def_dev_name = get_default_device_name(domain);
+	non_sec_dev_name = get_non_secure_device_name(domain);
```
quic-parihar: why not fetch both together in a single function

src/adsp_default_listener.c
quic-vkatoch: Sure, will check

src/adsp_default_listener.c
quic-vkatoch: Updated

src/fastrpc_apps_user.c
```diff
@@ -3217,94 +3217,72 @@ static void domain_deinit(int domain) {
...
 }
 
 int open_device_node(int domain_id) {
   int dev = -1, nErr = 0;
   int domain = GET_DOMAIN_FROM_EFFEC_DOMAIN_ID(domain_id);
   int sess_id = GET_SESSION_ID_FROM_DOMAIN_ID(domain_id);
+  const char *secure_dev = NULL, *non_secure_dev = NULL;
 
   switch (domain) {
   case ADSP_DOMAIN_ID:
```
ekanshibu: is switch-case really needed now? can't you just check if it is_valid_domain()?

src/adsp_default_listener.c
```diff
@@ -227,11 +140,25 @@ static int fastrpc_wait_for_device(int domain)
 		 /* Loop over all events in the buffer. */
 		for (char *ptr = buffer; ptr < buffer + len;
 					ptr += sizeof(struct inotify_event) + event->len) {
+			const char *sec_dev_basename = NULL, *non_sec_dev_basename = NULL;
 			event = (struct inotify_event *) ptr;
+
+			/* Extract just the device name from full path for comparison */
+			if (sec_dev_name)
```
ekanshibu: isn't sec_dev_name and non_sec_dev_name will always be populated here?

quic-vkatoch: Yes, will update

src/fastrpc_apps_user.c
quic-vkatoch: Ack, will update

quic-parihar (approved):

ekanshibu:
pls resolve merge conflicts

inc/fastrpc_internal.h
```diff
@@ -515,8 +524,6 @@ int ioctl_optimization(int dev, uint32_t max_concurrency);
 int ioctl_mdctx_manage(int dev, int req, void *user_ctx,
 	unsigned int *domain_ids, unsigned int num_domain_ids, uint64_t *ctx);
 
-const char* get_secure_domain_name(int domain_id);
-
-#include "fastrpc_ioctl.h"
+int is_async_fastrpc_supported(void);
```
ekanshibu: why?

src/fastrpc_apps_user.c
```diff
@@ -3029,108 +3029,76 @@ static void domain_deinit(int domain) {
...
+      dev = open(non_secure_dev, O_NONBLOCK);
       if ((dev < 0) && ((errno == ENOENT) || (errno == EACCES))) {
-        // Open the default device node if actual device node
-        // is not present
         FARF(RUNTIME_RPC_HIGH,
-             "Device node %s open failed for domain %d (errno %s),"
-             "falling back to node %s \n",
-             get_domain_name(domain), domain, strerror(errno), DEFAULT_DEVICE);
-        dev = open(DEFAULT_DEVICE, O_NONBLOCK);
+             "Device node %s open failed for domain %d (errno %s)\n",
```
ekanshibu: is this if() really needed?

inc/fastrpc_internal.h
quic-vkatoch: will drop this

src/fastrpc_apps_user.c
quic-vkatoch: The non_secure_dev errno check only controls intermediate logging and does not affect return flow, final failure is already handled by the trailing if (dev < 0).

should drop it?

ekanshibu: yes, I don't see any reason to have this, in final error case we are anyway logging all the details

ekanshibu (approved):

Chennak-quic (approved):

PR #311, opened 2026-03-06
files: Fix service startup race condition with udev rules

### LINKED ISSUE #309: systemd service definitions are flawed
### FastRPC version

v1.0.3

### DSP firmware version

N/A

### SoC and device name

N/A

### Kernel version

N/A

### What happened?

As introduced in https://github.com/qualcomm/fastrpc/commit/97b649d1fbe9a14964a5ad56ea488791b3b4d418

The `After=` clause has no effect because it only affects ordering within a single set of service starts. `WantedBy=multiuser.target` then races the kernel module load. If it is reached before the kernel module is loaded, the daemon will never start.

I believe the correct way of doing this is to set `WantedBy=` on the device unit. But since there are two device units that don't necessarily exist at same the time the service unit is loaded, it should be done by `Wants=` from the device unit, which needs inserting into the udev rule instead. I already suggested this in #245.

### Steps to reproduce

N/A

### Relevant log output

```shell

```

quic-vkatoch:
The After= clause has no effect on the race between WantedBy=multi-user.target and kernel module loading. If multi-user.target is reached before the kernel module loads, the daemon will never start.

Fix by using ENV{SYSTEMD_WANTS} in udev rules to trigger services when FastRPC devices are detected.

Changes:
- Add 71-fastrpc.rules with device-triggered service startup
- Remove [Install] sections from all service files
- Update configure.ac to use pkg-config to query udevdir variable
- Update files/Makefile.am to install udev rules

Services now start only when device nodes are created, eliminating the race condition.

Fixes: #309

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -39,6 +39,9 @@ AM_PROG_CC_C_O
 
 # Checks for libraries.
 
+# Enable pkg-config
+PKG_PROG_PKG_CONFIG
+
 # Configure systemd unit directory option (--with-systemdsystemunitdir)
 AC_ARG_WITH([systemdsystemunitdir],
   [AS_HELP_STRING([--with-systemdsystemunitdir=DIR],
@@ -48,12 +51,23 @@ AC_ARG_WITH([systemdsystemunitdir],
 AC_MSG_NOTICE([systemd unit directory: $systemdsystemunitdir])
 AC_SUBST([systemdsystemunitdir])
 
+# Configure udev rules directory using pkg-config
+AC_ARG_WITH([udevrulesdir],
+  [AS_HELP_STRING([--with-udevrulesdir=DIR],
+                  [Directory for udev rules files])],
+  [udevrulesdir="$withval"],
+  [
+    # Get udev directory from pkg-config
+    PKG_CHECK_VAR([udevrulesdir], [udev], [udevdir],
+      [udevrulesdir="${udevrulesdir}/rules.d"],
+      [AC_MSG_ERROR([Could not determine udev rules directory. Please install udev development files or use --with-udevrulesdir=DIR])])
+  ])
+AC_MSG_NOTICE([udev rules directory: $udevrulesdir])
+AC_SUBST([udevrulesdir])
+
 # Checks for typedefs, structures, and compiler characteristics.
 
 # Checks for library functions.
-
-# Enable pkg-config
-PKG_PROG_PKG_CONFIG
 
 # Check for libyaml only if not Android
 AS_IF([test "$compile_for_android" = no], [
--- a/files/71-fastrpc.rules
+++ b/files/71-fastrpc.rules
@@ -1,0 +1,28 @@
+# udev rules for Qualcomm FastRPC devices
+# These rules set permissions and systemd tags for FastRPC device nodes
+
+# aDSP devices
+ACTION=="add", SUBSYSTEM=="misc", KERNEL=="fastrpc-adsp", GROUP="render", MODE="0666", TAG+="systemd", ENV{SYSTEMD_WANTS}+="adsprpcd.service"
+ACTION=="add", SUBSYSTEM=="misc", KERNEL=="fastrpc-adsp-secure", TAG+="systemd", ENV{SYSTEMD_WANTS}+="adsprpcd.service"
+ACTION=="add", SUBSYSTEM=="misc", KERNEL=="fastrpc-adsp", GROUP="render", MODE="0666", TAG+="systemd", ENV{SYSTEMD_WANTS}+="adsprpcd_audiopd.service"
+ACTION=="add", SUBSYSTEM=="misc", KERNEL=="fastrpc-adsp-secure", TAG+="systemd", ENV{SYSTEMD_WANTS}+="adsprpcd_audiopd.service"
+
+# cDSP devices
+ACTION=="add", SUBSYSTEM=="misc", KERNEL=="fastrpc-cdsp", GROUP="render", MODE="0666", TAG+="systemd", ENV{SYSTEMD_WANTS}+="cdsprpcd.service"
+ACTION=="add", SUBSYSTEM=="misc", KERNEL=="fastrpc-cdsp-secure", TAG+="systemd", ENV{SYSTEMD_WANTS}+="cdsprpcd.service"
+
+# cDSP1 devices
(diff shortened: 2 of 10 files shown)


configure.ac
```diff
@@ -48,6 +48,15 @@ AC_ARG_WITH([systemdsystemunitdir],
...
 AC_SUBST([systemdsystemunitdir])
 
+# Configure udev rules directory option (--with-udevrulesdir)
+AC_ARG_WITH([udevrulesdir],
+  [AS_HELP_STRING([--with-udevrulesdir=DIR],
+                  [Directory for udev rules files (default: /lib/udev/rules.d)])],
+  [udevrulesdir="$withval"],
+  [udevrulesdir="/lib/udev/rules.d"])
+AC_MSG_NOTICE([udev rules directory: $udevrulesdir])
+AC_SUBST([udevrulesdir])
```
basak-qcom, line 66: Can you use pkg-config instead, querying for the `udevdir` variable? On Debian, `/usr/lib/udev/rules.d` is used (due to usrmerge), and I'd prefer for that to be automatic rather than have to use this project-specific `--with-udevrulesdir` or have the fallback diversion.

basak-qcom (changes requested):
The change itself looks good to me. Thanks! One comment inline on the autoconf change.

configure.ac
quic-vkatoch: Updated

basak-qcom: It says "Get udev directory from pkg-config" but doesn't seem to be?

quic-vkatoch: So now the `--with-udevrulesdir=DIR` option is available, so if another distro wants to install the rules into a specific path, they can provide it explicitly. If we do not specify anything, the directory is taken from pkg-config. I just want to confirm that I am understanding this correctly.

Here is what I observe during Yocto compilation:

Image

quic-vkatoch: @basak-qcom, can you please check once?

basak-qcom: Sorry I missed your reply before.

I understand that `--with-udevrulesdir=DIR` is available. But what I mean is that regardless of what distro you're building on, it shouldn't be necessary to use this option because `pkg-config --variable=udevdir udev` should provide the correct per-distro directory to use, and then autoconf should use that one automatically. It's fine to have an override, but it should never be necessary to use this.

quic-vkatoch:
Raised a PR [#1879](https://github.com/qualcomm-linux/meta-qcom/pull/1879) in the meta‑qcom repo as the requirement was to use the render group.

quic-vkatoch:
@basak-qcom, could you please review this?

files/71-fastrpc.rules
```diff
@@ -0,0 +1,28 @@
+# udev rules for Qualcomm FastRPC devices
+# These rules set permissions and systemd tags for FastRPC device nodes
+
+# aDSP devices
+ACTION=="add", SUBSYSTEM=="misc", KERNEL=="fastrpc-adsp", GROUP="render", MODE="0660", TAG+="systemd", ENV{SYSTEMD_WANTS}+="adsprpcd.service"
+ACTION=="add", SUBSYSTEM=="misc", KERNEL=="fastrpc-adsp-secure", GROUP="render", MODE="0660", TAG+="systemd", ENV{SYSTEMD_WANTS}+="adsprpcd.service"
```
ekanshibu: I'm still not sure about the permissions. I did a quick check on IQ8 device and I see the following permissions for dri devices:
crw-rw-rw-  1 root render 226, 128 Jan  1 00:00 renderD128
crw-rw----  1 root video  226,   0 Jan  1 00:00 card0

here render device carries 666 permissions. I think we should have the same here also. 

@lumag, @basak-qcom would need your suggestions here.

files/71-fastrpc.rules
```diff
@@ -0,0 +1,28 @@
...
+ACTION=="add", SUBSYSTEM=="misc", KERNEL=="fastrpc-adsp", GROUP="render", MODE="0660", TAG+="systemd", ENV{SYSTEMD_WANTS}+="adsprpcd_audiopd.service"
+ACTION=="add", SUBSYSTEM=="misc", KERNEL=="fastrpc-adsp-secure", GROUP="render", MODE="0660", TAG+="systemd", ENV{SYSTEMD_WANTS}+="adsprpcd_audiopd.service"
+
+# cDSP devices
+ACTION=="add", SUBSYSTEM=="misc", KERNEL=="fastrpc-cdsp", GROUP="render", MODE="0660", TAG+="systemd", ENV{SYSTEMD_WANTS}+="cdsprpcd.service"
+ACTION=="add", SUBSYSTEM=="misc", KERNEL=="fastrpc-cdsp-secure", GROUP="render", MODE="0660", TAG+="systemd", ENV{SYSTEMD_WANTS}+="cdsprpcd.service"
```
ekanshibu: @quic-vkatoch let's keep GROUP, MODE of secure devices unchanged, as only daemons should use these devices.

quic-vkatoch: Ack, will update

files/71-fastrpc.rules
basak-qcom: This relates to https://github.com/qualcomm-linux/pkg-fastrpc/issues/16

But it seems to me that 0666 is surprising. I'd expect 0660 and then proper use of the group, whatever it ends up being. If we need 0666, I'd like to see an explanation of why.

quic-vkatoch: On meta‑qcom flashed builds, the render group is showing the mode as 0666. I believe this behavior is picked up from the upstream configuration.
The reason 0666 appears here is that the upstream system explicitly supports world‑writable DRM render nodes. This allows users to avoid modifying their existing setup for applications and continue using them in the same way.
**Reference:**
Here is the link: https://github.com/systemd/systemd/blob/main/meson_options.txt#L346C9-L346C26

basak-qcom: Thanks! I found the upstream discussion here:

https://github.com/systemd/systemd/pull/7112

The key reason for this default was:

> What matters here really is that the stuff is designed to be safe to be made world writable, modulo any software bugs.

I think this logic is reasonable, but how does it apply to the fastrpc device nodes? For example, what about resource exhaustion: could that be done by a non-root user (eg. `nobody`) and would that be acceptable?

quic-vkatoch:
Closing this PR, it has been superseded by PR #333.

PR #317, opened 2026-03-18
Qualcomm Preflight Checks: Replace pull_request_target with pull_request

mynameistechno:
Running untrusted code on the pull_request_target trigger may lead to security vulnerabilities. These vulnerabilities include cache poisoning and granting unintended access to write privileges or secrets. 

https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target

We should update all usage of pull_request_target in all workflow files.


ekanshibu:
This looks to be done now with PR:https://github.com/qualcomm/fastrpc/pull/306

PR #319, opened 2026-03-24
apps_mem: fallback to legacy mmap/unmap when internal API fails (Talos)

Jianping-Li:
On Talos, fastrpc_mmap_internal/munmap_internal can fail or be unsupported. Fallback to fastrpc_mmap and then remote_mmap64_internal, and use remote_munmap64 when internal munmap fails.

### DIFF
--- a/src/apps_mem_imp.c
+++ b/src/apps_mem_imp.c
@@ -23,6 +23,7 @@
 #include <string.h>
 #include <sys/mman.h>
 #include "fastrpc_hash_table.h"
+#include "fastrpc_internal.h"
 
 #define ADSP_MMAP_HEAP_ADDR 4
 #define ADSP_MMAP_REMOTE_HEAP_ADDR 8
@@ -47,6 +48,7 @@ struct mem_info {
   int32_t size;
   int32_t mapped;
   uint32_t rflags;
+  enum fastrpc_map_type map_method;
 };
 
 /* Delete and free all nodes in hash-table */
@@ -167,9 +169,19 @@ __QAIC_IMPL(apps_mem_request_map64)(int heapid, uint32_t lflags, uint32_t rflags
             AEE_ENORPCMEMORY);
     fd = rpcmem_to_fd_internal(buf);
     VERIFYC(fd > 0, AEE_EBADPARM);
-    VERIFY(AEE_SUCCESS ==
-           (nErr = fastrpc_mmap_internal(domain, fd, buf, 0, len,
-                                         rflags, vadsp)));
+    nErr = fastrpc_mmap_internal(domain, fd, buf, 0, len,
+                                          rflags, vadsp);
+    minfo->map_method = MEM_MAP;
+    if (nErr == AEE_EUNSUPPORTED) {
+      rpcmem_free_internal(buf);
+      buf = NULL;
+      fd = -1;
+      minfo->map_method = MMAP_64;
+      VERIFY(AEE_SUCCESS == (nErr = remote_mmap64_internal(fd, rflags, 
+                              (uint64_t)buf, len, (uint64_t *)vadsp)));
+    } else if (nErr) {
+          goto bail;
+    }
     pbuf = (uint64_t)buf;
     *vapps = pbuf;
     minfo->vapps = *vapps;
@@ -256,11 +268,21 @@ __QAIC_IMPL(apps_mem_request_unmap64)(uint64_t vadsp,
        * So continue to unmap to avoid driver global maps leak.
        * For other cases, use fastrpc_munmap_internal as they were mapped with fastrpc_mmap_internal.
        */
-      if (mfree) {
-         VERIFY(AEE_SUCCESS == (nErr = fastrpc_munmap_internal(domain, (uint64_t)vadsp, len)));
-      } else {
-         VERIFY(AEE_SUCCESS == (nErr = remote_munmap64((uint64_t)vadsp, len)));
+      switch (minfo->map_method) {
+          case MEM_MAP:
+              VERIFY(AEE_SUCCESS ==
+                    (nErr = fastrpc_munmap_internal(domain,
+                                                    (uint64_t)vadsp, len)));
+              break;
+          case MMAP_64:
+              VERIFY(AEE_SUCCESS ==
+                    (nErr = remote_munmap64((uint64_t)vadsp, len)));
+              break;
+          default:
+              VERIFYC(0, AEE_EBADPARM);
+              break;
       }
+
       if (!mfree)
          goto bail;
    }


src/apps_mem_imp.c
```diff
@@ -167,12 +167,18 @@ __QAIC_IMPL(apps_mem_request_map64)(int heapid, uint32_t lflags, uint32_t rflags
             AEE_ENORPCMEMORY);
     fd = rpcmem_to_fd_internal(buf);
     VERIFYC(fd > 0, AEE_EBADPARM);
-    VERIFY(AEE_SUCCESS ==
-           (nErr = fastrpc_mmap_internal(domain, fd, buf, 0, len,
-                                         rflags, vadsp)));
-    pbuf = (uint64_t)buf;
-    *vapps = pbuf;
-    minfo->vapps = *vapps;
```
quic-vkatoch, line 175: Shouldn’t we be storing these addresses?

Jianping-Li: Thank you for the reminder. I accidentally deleted these three lines because they did not affect the Talos test results, but considering compatibility, they should be added back.

src/apps_mem_imp.c
```diff
@@ -167,13 +167,19 @@ __QAIC_IMPL(apps_mem_request_map64)(int heapid, uint32_t lflags, uint32_t rflags
...
+    nErr = fastrpc_mmap_internal(domain, fd, buf, 0, len,
+                                          rflags, vadsp);   
+    if (nErr != AEE_SUCCESS) {
+      rpcmem_free_internal(buf);
+      buf = NULL;
+      fd = -1;
+      VERIFY(AEE_SUCCESS == (nErr = remote_mmap64_internal(fd, rflags,
```
Chennak-quic, line 180: Falling back to remote_mmap64_internal() for all error cases of fastrpc_mmap_internal() is incorrect. A fallback is acceptable only when the DSP firmware does not support the ioctl MEM_MAP interface (adsp_process_group_mem_map() on DSP). In all other cases, the function should return an error

Chennak-quic (changes requested):

src/apps_mem_imp.c
```diff
@@ -167,13 +167,19 @@ __QAIC_IMPL(apps_mem_request_map64)(int heapid, uint32_t lflags, uint32_t rflags
...
-    minfo->vapps = *vapps;
+    nErr = fastrpc_mmap_internal(domain, fd, buf, 0, len,
+                                          rflags, vadsp);   
+    if (nErr == AEE_EUNSUPPORTED) {
+      rpcmem_free_internal(buf);
+      buf = NULL;
+      fd = -1;
+      VERIFY(AEE_SUCCESS == (nErr = remote_mmap64_internal(fd, rflags, 
+                              (uint64_t)buf, len, (uint64_t *)vadsp)));
+    }
```
ekanshibu, line 184: else {
  goto bail;
}

src/apps_mem_imp.c
```diff
@@ -257,7 +263,10 @@ __QAIC_IMPL(apps_mem_request_unmap64)(uint64_t vadsp,
        * For other cases, use fastrpc_munmap_internal as they were mapped with fastrpc_mmap_internal.
        */
       if (mfree) {
-         VERIFY(AEE_SUCCESS == (nErr = fastrpc_munmap_internal(domain, (uint64_t)vadsp, len)));
+         nErr = fastrpc_munmap_internal(domain, (uint64_t)vadsp, len);
+         if (nErr == AEE_EUNSUPPORTED) {
+            VERIFY(AEE_SUCCESS == (nErr = remote_munmap64((uint64_t)vadsp, len)));
+         }
```
ekanshibu: else {
  goto bail;
}

ekanshibu (changes requested):

src/apps_mem_imp.c
```diff
@@ -167,13 +167,19 @@ __QAIC_IMPL(apps_mem_request_map64)(int heapid, uint32_t lflags, uint32_t rflags
...
+      rpcmem_free_internal(buf);
+      buf = NULL;
+      fd = -1;
+      VERIFY(AEE_SUCCESS == (nErr = remote_mmap64_internal(fd, rflags, 
+                              (uint64_t)buf, len, (uint64_t *)vadsp)));
+    }
   }
+  pbuf = (uint64_t)buf;
+  *vapps = pbuf;
+  minfo->vapps = *vapps;
```
quic-vkatoch: Keep it in the else block as before

src/apps_mem_imp.c
```diff
@@ -257,7 +265,12 @@ __QAIC_IMPL(apps_mem_request_unmap64)(uint64_t vadsp,
        * For other cases, use fastrpc_munmap_internal as they were mapped with fastrpc_mmap_internal.
        */
       if (mfree) {
-         VERIFY(AEE_SUCCESS == (nErr = fastrpc_munmap_internal(domain, (uint64_t)vadsp, len)));
+         nErr = fastrpc_munmap_internal(domain, (uint64_t)vadsp, len);
```
Chennak-quic: if fastrpc_mmap_internal() is not supported, the user library ends up being mapped via remote_mmap_internal(). In that case, why does the unmap path still attempt fastrpc_munmap_internal() first? We should record which mmap path was actually used (for example in minfo or a similar structure) and then make the munmap decision based on that state

Chennak-quic (changes requested):

ekanshibu (dismissed):

quic-vkatoch:
Please update the commit message.
Change LGTM.

src/apps_mem_imp.c
```diff
@@ -256,10 +267,10 @@ __QAIC_IMPL(apps_mem_request_unmap64)(uint64_t vadsp,
        * So continue to unmap to avoid driver global maps leak.
        * For other cases, use fastrpc_munmap_internal as they were mapped with fastrpc_mmap_internal.
        */
-      if (mfree) {
-         VERIFY(AEE_SUCCESS == (nErr = fastrpc_munmap_internal(domain, (uint64_t)vadsp, len)));
+      if (mfree && minfo->map_method) {
+        VERIFY(AEE_SUCCESS == (fastrpc_munmap_internal(domain, (uint64_t)vadsp, len)));
```
quic-vkatoch: Please add the nErr assignment

src/apps_mem_imp.c
```diff
@@ -47,6 +47,7 @@ struct mem_info {
   int32_t size;
   int32_t mapped;
   uint32_t rflags;
+  bool map_method;
```
Chennak-quic: how about having an enum for map_method though we have only two methods as of now.  @ekanshibu , what do you suggest ?

ekanshibu: I think that would be good to have, we can stick with something similar to QDA patch:
FASTRPC_MAP_REQUEST_LEGACY
FASTRPC_MAP_REQUEST_ATTR

ekanshibu (dismissed):

src/apps_mem_imp.c
Chennak-quic: can we reuse existing enum fastrpc_map_type defined in  https://github.com/qualcomm/fastrpc/blob/development/inc/fastrpc_internal.h ?

anankulk:
Submitted the jobs manually with updated working commands : 

qcs6490 (signed PD) : [LAVA | Scheduler | Jobs | 73742](https://lava-oss.qualcomm.com/scheduler/job/73742)   [Passed]
qcs9100-ride (signed PD) : [LAVA | Scheduler | Jobs | 73746](https://lava-oss.qualcomm.com/scheduler/job/73746) [Passed]
qcs8300-ride (signed) : [LAVA | Scheduler | Jobs | 73747](https://lava-oss.qualcomm.com/scheduler/job/73747) [Passed]

ekanshibu (approved):

src/apps_mem_imp.c
```diff
@@ -256,11 +268,21 @@ __QAIC_IMPL(apps_mem_request_unmap64)(uint64_t vadsp,
        * So continue to unmap to avoid driver global maps leak.
        * For other cases, use fastrpc_munmap_internal as they were mapped with fastrpc_mmap_internal.
        */
-      if (mfree) {
-         VERIFY(AEE_SUCCESS == (nErr = fastrpc_munmap_internal(domain, (uint64_t)vadsp, len)));
-      } else {
-         VERIFY(AEE_SUCCESS == (nErr = remote_munmap64((uint64_t)vadsp, len)));
+      switch (minfo->map_method) {
```
Chennak-quic, line 271: Don't we need to check for mfree NULL case now ?

ekanshibu:
This issue is addressed in https://github.com/qualcomm/fastrpc/pull/338

PR #320, opened 2026-03-26
Add MACHINE_NAME evn variable support

Srinivas-Kandagatla:
Add MACHINE_NAME enviroment variable as fallback mechanism to detect
    machine name, this should allow setups which will not have
    access to /sys/firmware/ entries, example containers itself to be able
    to use fastrpc libraries and dynamically get hold of hexagon binary
    paths.

Apart from this change the file also has some LR chars, this pr converts them before adding the MACHINE_NAME support


src/fastrpc_config_parser.c
```diff
@@ -169,13 +169,20 @@ void configure_dsp_paths() {
...
+    if (fgets(machine_name, sizeof(machine_name), file) != NULL)
+        // Remove trailing newline if present
+        machine_name[strcspn(machine_name, "\n")] = '\0';
 
-  fclose(file);
-  parse_config_dir(machine_name);
+    fclose(file);
+    parse_config_dir(machine_name);
+  } else {
+    char *env_value = getenv("MACHINE_NAME");
```
ekanshibu, line 180: Changes looks good to me, just 1 basic question. If the plan is to keep containers platform agnostic, how and when will the environment variable get set? Is it expected from the user to set the env before running their use-case?

Srinivas-Kandagatla: Users will be able to deploy containers like this as example, we could do exactly same in docker compose :
` docker run   -e MACHINE_NAME="$(tr -d '\0' < /sys/firmware/devicetree/base/model)"  IMAGE`
This is pertty much generic and will work for all the platforms.

ekanshibu (approved):

src/fastrpc_config_parser.c
```diff
@@ -169,13 +169,20 @@ void configure_dsp_paths() {
...
+    // fallback to look for MACHINE_NAME in environment variable
+    if (env_value) {
+      snprintf(machine_name, sizeof(machine_name), "%s", env_value);
+      machine_name[strcspn(machine_name, "\n")] = '\0';
+      parse_config_dir(machine_name);
```
Chennak-quic, line 185: Consider moving parse_config_dir() call outside the if and else cases, since it’s common to both

Chennak-quic (changes requested):

basak-qcom:
This is fine from a Debian packaging perspective as the environment variable won't be required, and on a host system it will just work. I wonder if we should just expect containers to bind mount the right part of `/sys/firmware` though. @koenkooi do you have an opinion, please?

koenkooi:
> This is fine from a Debian packaging perspective as the environment variable won't be required, and on a host system it will just work. I wonder if we should just expect containers to bind mount the right part of `/sys/firmware` though. @koenkooi do you have an opinion, please?

We considered this, but too many[^1] container runtimes filter access to `/sys` and `/proc`, even when bind-mounted to a different place. So a different method is needed, and as you point out, it's a fallback that doesn't affect 'proper' systems.

[^1]: I'd say "not enough", but people seem to like container runtimes that are bad the actual _containing_ portion.

basak-qcom:
Thanks! In that case this approach looks fine from my perpsective.

src/fastrpc_config_parser.c
Srinivas-Kandagatla: Nope, parse_config_dir is conditional if MACHINE_NAME env is available, and its conditional in case of file aswell, so it does not make sense to move it to common. This will make the code look ugly if we try to do so.

Chennak-quic (approved):

PR #324, opened 2026-04-07
Add support for rpcmem_alloc2 

quic-anane:
## Support for Large Shared Memory Allocations

This PR introduces a new API:

- `rpcmem_alloc2`

The new API supports shared memory allocations **≥ 2GB**.

The existing `rpcmem_alloc` API uses an `int`-sized length parameter, which can overflow or fail for large allocations.  
`rpcmem_alloc2` avoids signed size limitations and safely supports large buffer requests.

### DIFF
--- a/src/rpcmem_linux.c
+++ b/src/rpcmem_linux.c
@@ -240,6 +240,10 @@ void *rpcmem_alloc(int heapid, uint32_t flags, int size) {
   return rpcmem_alloc_internal(heapid, flags, size);
 }
 
+void *rpcmem_alloc2(int heapid, uint32_t flags, size_t size) {
+  return rpcmem_alloc_internal(heapid, flags, (size_t)size);
+}
+
 void rpcmem_deinit_internal() { rpcmem_deinit(); }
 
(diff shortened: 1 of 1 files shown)


inc/remote.h
```diff
@@ -700,7 +700,16 @@ enum fastrpc_map_flags {
...
     FASTRPC_MAP_FD_NOMAP,
 
-    /** Update FASTRPC_MAP_MAX when adding new value to this enum **/
+    /**
+     * These 2 flags work the same way as FASTRPC_MAP_FD and
+     * FASTRPC_MAP_FD_DELAYED with the only difference being the DSP mapping
+     * created in the extended address space.
+     */
+    FASTRPC_MAP_FD_EXTENDED,
+    FASTRPC_MAP_FD_DELAYED_EXTENDED,
```
ekanshibu: should we bring this separately? What would you suggest, @Chennak-quic?

ekanshibu (approved):

Chennak-quic (approved):

PR #337, opened 2026-04-22
Fix missing unlock on VERIFY failure paths

0222826398:
Fix two paths where VERIFY could jump to bail while a mutex was still held. Always release the mutex on error paths to prevent deadlock during cleanup.


quic-vkatoch (approved):
lgtm

ekanshibu (approved):

PR #341, opened 2026-04-24
ci: Switch workflows to pull_request event

ekanshibu:
Replace pull_request_target with pull_request in PR workflows to follow GitHub security practices and avoid elevated permissions during PR validation.

### DIFF
--- a/.github/workflows/pkg-build-pr-check.yml
+++ b/.github/workflows/pkg-build-pr-check.yml
@@ -35,5 +35,3 @@ jobs:
       upstream-repo-ref: ${{ github.event.pull_request.head.ref }}
       pkg-repo: ${{vars.PKG_REPO_GITHUB_NAME}}
       pr-number: ${{github.event.pull_request.number}}
-    secrets:
-      TOKEN: ${{ secrets.DEB_PKG_BOT_CI_TOKEN }}
--- a/.github/workflows/pkg-build-pr-check.yml
+++ b/.github/workflows/pkg-build-pr-check.yml
@@ -11,7 +11,7 @@ description: |
   a full build of the package if it were to include these changes.
 
 on:
-  pull_request_target:
+  pull_request:
     branches: [ main, development ]
     types:
       - ready_for_review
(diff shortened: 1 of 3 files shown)


njjetha:
moving from pull_request_target to pull_request would not have access to secrets you need to refactor workflow
https://github.qualcomm.com/pages/osdo/handbook/qcom-github/docs/github-actions-pull-request-target

.github/workflows/pkg-build-pr-check.yml
njjetha: It will not have access to secrets.

ekanshibu: @simonbeaudoin0935 based on the recent guidelines to move to pull_request from pull_Request_target do you have any plans to change this workflow?

simonbeaudoin0935: @ekanshibu Simply remove the line 38-39, the service bot secret is not used anymore in this case!

anankulk (approved):

Chennak-quic (approved):

mynameistechno (approved):

.github/workflows/pre_merge.yml
njjetha, line 31: For `pull_request` events especially those originating from forks secrets are not exposed regardless of secrets: inherit, so this setting has no effect on fork‑based PR runs.
However, for pull requests raised within the same repository (non‑forked PRs), secrets are available, and in that case secrets: inherit does make sense. It is also relevant for push and workflow_dispatch events.
I checked that reusable workflow don't use the secret

njjetha (approved):

.github/workflows/pre_merge.yml
```diff
@@ -15,7 +15,7 @@ on:
     - 'main'
     - 'development'
   workflow_dispatch:
-  pull_request_target: # requiring access to base repo
+  pull_request: # requiring access to base repo
```
quic-mtharu, line 18: Remove the comment if it is changed back to pull_request.

PR #372, opened 2026-08-07
udev: start DSP remoteprocs left offline

### LINKED ISSUE #371: Regression with fastrpc 1.0.6 on a few platforms
Hi

Arun Kumar pinged me about regressions running fastrpc_tests on glymur-crd with 1.0.6 which he thinks could be permission related. I don't have a glymur-crd handy, but I tried running qcom-linux-testkit tests on platforms available in LAVA, first with latest qcom-next builds which have 1.0.6, then with an older build from Jul 30th which had version 1.0.2.

  | Build (qcom-deb-images) | Board                        | ADSP Signed | CDSP Signed | CDSP Unsigned | Job |
  |--------------------------|-------------------------------|:-----------:|:-----------:|:--------------:|-----|
  | 2026-07-30 (`c6ae6fc`)   | qcs6490-rb3gen2-vision-kit    | PASS (3/3)  | PASS (3/3)  | PASS (3/3)      | [350836](https://lava.infra.foundries.io/scheduler/job/350836) |
  | 2026-08-05 (`70c4895`)   | qcs6490-rb3gen2-vision-kit    | PASS (3/3)  | **FAIL (0/3)** | **FAIL (0/3)** | [350580](https://lava.infra.foundries.io/scheduler/job/350580) |

Failure signature (both CDSP Signed and CDSP Unsigned invocations, 2026-08-05 build):

  ERROR 0x72: Unable to create FastRPC session on domain 3
  ERROR 0x72: Failed to compute sum on domain 3
  ERROR 0x72: Failed to find max on domain 3
  hap_example function FAILED
  [FAIL] libhap_example.so (error code: 0x72)

  `remoteproc` state at time of failure:

  rproc.adsp: remoteproc1 state=running fw=qcom/qcs6490/adsp.mbn name=adsp
  rproc.cdsp: remoteproc3 state=offline  fw=qcom/qcs6490/cdsp.mbn name=cdsp

  `rproc.cdsp` reports `state=offline`, consistent with the CDSP FastRPC session
  creation failure (error 0x72) above.

## Boards that ran cleanly (2026-07-30 build, `c6ae6fc`)

  For comparison, all domains passed (9/9 subtests) on these boards/jobs:

  | Board            | Job |
  |-------------------|-----|
  | qcs615-ride       | [350834](https://lava.infra.foundries.io/scheduler/job/350834) |
  | monaco-evk        | [350835](https://lava.infra.foundries.io/scheduler/job/350835) |
  | qcs6490-rb3gen2-vision-kit | [350836](https://lava.infra.foundries.io/scheduler/job/350836) |
  | qcs8300-ride      | [350837](https://lava.infra.foundries.io/scheduler/job/350837) |

(glymur-crd and lemans-evk are omitted from this report as those runs did not
  execute the test on this build cycle: glymur-crd's test was skipped by SoC
  blacklist, and lemans-evk's job hit an unrelated infra failure, login-action
  timeout.)

lool:
When qcom_q6v5_pas is loaded from the initramfs, rproc_add()
auto-boots the core before the root filesystem is mounted.
/lib/firmware is not reachable yet, so request_firmware() fails with
-ENOENT, and the remoteproc core has no retry path: the DSP stays
"offline" for the rest of the boot even though the firmware is
present on disk. With no DSP there is no rpmsg channel, so no
/dev/fastrpc-* node appears and none of the daemons in
60-fastrpc.rules can start.

Add a udev rule that writes "start" to the remoteproc state attribute.
The event fires twice: once in the initramfs, where the write still fails
for the same reason the kernel's did, and again from the coldplug pass
that systemd-udev-trigger.service runs after switch_root, where the
firmware is reachable and the boot succeeds. udev logs and ignores the
failed write, so the first pass is harmless.

Only cores that qcom_q6v5_pas auto-boots and that FastRPC uses are
matched. modem and wpss have .auto_boot = false: modem is absent by
design on some SoCs, and ath11k owns the wpss lifecycle.

The ATTR{state}=="offline" gate is required: rproc_boot() increments
rproc->power unconditionally, so writing to an already-running core would
pin an extra reference and prevent a later shutdown or subsystem restart
from taking effect.

This is a userspace stopgap for the kernel bug; it can be dropped once a
kernel-side retry lands.

Relates to https://github.com/qualcomm-linux/kernel/issues/923
Fixes: #371

### DIFF
--- a/files/59-fastrpc-remoteproc.rules
+++ b/files/59-fastrpc-remoteproc.rules
@@ -1,0 +1,46 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Bring up DSP remote processors that the kernel left permanently offline.
+#
+# Workaround for https://github.com/qualcomm-linux/kernel/issues/923: when
+# qcom_q6v5_pas is loaded from the initramfs, rproc_add() auto-boots the core
+# before the root filesystem is mounted. /lib/firmware is not reachable yet, so
+# request_firmware() fails with -ENOENT, and the remoteproc core has no retry
+# path -- the DSP stays "offline" for the whole boot even though the firmware is
+# present on disk. Without a DSP there is no rpmsg channel, so no /dev/fastrpc-*
+# node is ever created and none of the daemons in 60-fastrpc.rules can start.
+#
+# Numbered below 60-fastrpc.rules to reflect that a running DSP is a
+# precondition for the FastRPC device nodes that rule set matches on.
+#
+# These events fire twice: once in the initramfs, where the write still fails
+# because the firmware remains unreachable, and again from the coldplug pass
+# that systemd-udev-trigger.service runs after switch_root, where it succeeds.
+# The first pass therefore logs
+#     Failed to write "start" to sysfs attribute "state", ignoring
+# which is expected and harmless -- udev ignores a failed ATTR{} write, so the
+# event and the boot both complete. A genuinely missing firmware blob logs the
+# same message, so that line on its own does not indicate a problem.
+#
+# Only cores that qcom_q6v5_pas auto-boots and that FastRPC actually uses are
+# matched.
+#
+# ATTR{name} is the remoteproc name, which for qcom_q6v5_pas comes from the
+# descriptor's .sysmon_name -- not from the FastRPC domain name. The sensor DSP
+# is "slpi" there; "sdsp" is only its FastRPC-side name (/dev/fastrpc-sdsp,
+# SDSP_DOMAIN_NAME), and no remoteproc is ever exposed under it.
+#
+# Deliberately not matched:
+#   - modem: .auto_boot = false on many SoCs, and some (e.g. qcs6490) ship no
+#     modem firmware at all, so it is offline by design.
+#   - wpss: .auto_boot = false, and it is booted on demand by ath11k, which
+#     owns its lifecycle.
+#   - lpaicp: auto-booted, but not a FastRPC domain.
+#
+# The ATTR{state}=="offline" gate is required: rproc_boot() increments
+# rproc->power unconditionally, so writing to an already-running core pins an
+# extra reference and prevents a later shutdown or subsystem restart from taking
+# effect. udev sorts all match tokens ahead of all assignments within a rule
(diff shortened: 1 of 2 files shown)


lool:
I tested this on kodiak; this is the failure mode:
https://lava.infra.foundries.io/scheduler/job/353543

I manually installed the new file in /etc/udev/rules.d/59-fastrpc-remoteproc-inline.rules, ran update-initramfs -u, rebooted, and remote cores worked. That part was local, I don't have a LAVA run for that.

files/59-fastrpc-remoteproc.rules
```diff
@@ -0,0 +1,38 @@
...
+#     qcs6490) ship no modem firmware at all, so it is offline by design.
+#   - wpss is excluded: it has .auto_boot = false and is booted on demand by
+#     ath11k, which owns its lifecycle.
+#
+# The ATTR{state}=="offline" gate is required: rproc_boot() increments
+# rproc->power unconditionally, so writing to an already-running core pins an
+# extra reference and prevents a later shutdown or subsystem restart from taking
+# effect. udev sorts all match tokens ahead of all assignments within a rule
+# line, so the gate is evaluated before the write even though both name state.
+ACTION=="add", SUBSYSTEM=="remoteproc", ATTR{name}=="adsp|cdsp|cdsp1|gpdsp0|gpdsp1|sdsp|slpi", ATTR{state}=="offline", ATTR{state}="start"
```
lumag: I think there is only one of slpi and sdsp.

lool: Thanks, good catch, sdsp is a fastrpc internal name; I've dropped it and kept slpi that was apparently used on rb3gen2 and updated the comment in the udev rules. Apparently qcom-next has a soccp as well, should that be here?

lool:
We discussed this with the FastRPC team:
- the proper fix is in the kernel, but that might be some time away
- workarounds can be delivered in initramfs generators, but they are also some time away; for instance for initramfs-tools: https://salsa.debian.org/kernel-team/initramfs-tools/-/merge_requests/201
- this workaround of adding udev rules to fastrpc is a bit of a kludge in that people will have a hard time to discover how to get their DSP working, and will need the fastrpc package installed (or a similar udev rule) to start their DSP after initramfs
- the FastRPC team is not keen to carry the workaround without an agreed timeline for the proper fix, but they are ok with distro patches carrying this change

PR #380, opened 2026-08-21
fix(dsp_check): correct DSP module search scoping and honor --with-config-base-dir

### LINKED ISSUE #385: /usr/bin/dsp_check hardcodes /usr/share/qcom
/usr/bin/dsp_check fails on Debian because it assumes /usr/share/qcom, which isn't generally true. It should honor CONFIG_BASE_DIR, but apparently does not.

Found in: 1.0.7.

Therefore I'm seeing the following output, which I believe is not what is expected:

```
$ ./dsp_check
DSP firmware path : /lib/firmware/
DSP build ID      : N/A
DSP dynamic libs path : N/A

DMA-BUF Heap Support:
System heap -> unavailable

DSP Summary:
DSP      State      SignedPD  UnsignedPD* FastRPC Support
---------------------------------------------------------------------------------
ADSP     Online     No        No          No (Missing DSP modules)
CDSP     Online     No        No          No (Missing DSP modules)

* If Unsigned PD not supported, needs signing to run
```

ekanshibu:
## Summary

Fixes two correctness bugs in `test/dsp_check.c` that could cause the
tool to report incorrect FastRPC support status, plus a minor build
hygiene fix.

### 1. Fix DSP module search path scoping (`build_search_for`)

`build_search_for()` enumerated **every** subdirectory under the
YAML-derived DSP library base path and added all of them to a DSP's
search list, instead of only adding the subdirectory matching that
specific DSP (e.g. `adsp/`, `cdsp/`, `sdsp/`). As a result, a DSP's
module search list could include other DSPs' library folders.

**Impact:** `dsp_check` could report `FastRPC Support: Yes` for a DSP
that doesn't actually have its own shell module installed, if another
DSP's folder happened to contain one.

**Fix:** Build the DSP-specific subdirectory path directly using the
DSP's own lower-cased name, consistent with the scoping already used
in `scan_yaml_shell_availability()` and `add_android_dsp_search_root()`.

### 2. Honor `--with-config-base-dir` instead of hardcoding `/usr/share/qcom`

`dsp_check.c` hardcoded `QCOM_BASE_DIR` as `/usr/share/qcom`, and two
functions additionally hardcoded the literal `/usr/share/qcom/conf.d`
path directly. This ignored the `--with-config-base-dir` configure
option already honored by the rest of the codebase (e.g.
`src/fastrpc_config_parser.c`'s `CONFIG_DIR`), which distros use to
relocate config files — for example, Debian's packaging passes
`--with-config-base-dir=/usr/share/hexagon-dsp`.

**Impact:** On such distros, `dsp_check` would keep looking under
`/usr/share/qcom/` and incorrectly report missing DSP modules/firmware
even though everything is correctly installed under the configured
path.

**Fix:**
- `test/dsp_check.c`: derive `QCOM_BASE_DIR` from the build-time
  `CONFIG_BASE_DIR` macro (falling back to `/usr/share/qcom/` if
  undefined), and build the `conf.d` path dynamically instead of
  hardcoding it twice.
- `test/Makefile.am`: pass `-DCONFIG_BASE_DIR` to `dsp_check_CFLAGS`,
  mirroring how `src/Makefile.am` already does for the FastRPC
  libraries.

### 3. Build hygiene: ignore compiled `test/dsp_check` binary

Added `test/dsp_check` to `.gitignore`, matching the existing entry
for `test/fastrpc_test`.

Fixes: https://github.com/qualcomm/fastrpc/issues/385
CRs-Fixed: 4658995

### DIFF
--- a/test/dsp_check.c
+++ b/test/dsp_check.c
@@ -27,7 +27,10 @@
 #define PATH_MAX 4096
 #endif
 
-#define QCOM_BASE_DIR "/usr/share/qcom"
+#ifndef CONFIG_BASE_DIR
+#define CONFIG_BASE_DIR "/usr/share/qcom/"
+#endif
+#define QCOM_BASE_DIR CONFIG_BASE_DIR
 #define MACHINE_MODEL_PATH "/sys/firmware/devicetree/base/model"
 
 #ifdef ANDROID
@@ -230,7 +233,11 @@ static char* find_dsp_library_path_from_yaml(void)
     if (read_text(MACHINE_MODEL_PATH, machine, sizeof(machine)) < 0)
         return NULL;
     machine[strcspn(machine, "\n")] = '\0';
-    DIR* dir = opendir("/usr/share/qcom/conf.d");
+
+    char conf_dir[PATH_MAX];
+    snprintf(conf_dir, sizeof(conf_dir), "%s/conf.d", QCOM_BASE_DIR);
+
+    DIR* dir = opendir(conf_dir);
     if (!dir)
         return NULL;
 
@@ -243,7 +250,7 @@ static char* find_dsp_library_path_from_yaml(void)
 
         char yaml_path[PATH_MAX];
         snprintf(yaml_path, sizeof(yaml_path),
-                 "/usr/share/qcom/conf.d/%s", de->d_name);
+                 "%s/%s", conf_dir, de->d_name);
 
         FILE* f = fopen(yaml_path, "r");
         if (!f)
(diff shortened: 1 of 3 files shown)


quic-vkatoch (approved):

test/dsp_check.c
```diff
@@ -230,7 +233,11 @@ static char* find_dsp_library_path_from_yaml(void)
     if (read_text(MACHINE_MODEL_PATH, machine, sizeof(machine)) < 0)
         return NULL;
     machine[strcspn(machine, "\n")] = '\0';
-    DIR* dir = opendir("/usr/share/qcom/conf.d");
+
+    char conf_dir[PATH_MAX];
+    snprintf(conf_dir, sizeof(conf_dir), "%s/conf.d", QCOM_BASE_DIR);
```
m-bhattar, line 238: Wouldn't this be /usr/share/qcom//conf.d? Can the extra "/" be removed?

ekanshibu: distro might or might not use "/", example:
https://github.com/qualcomm-linux/pkg-fastrpc/blob/debian/latest/debian/rules#L7

m-bhattar (approved):

quic-bkumar (approved):

Chennak-quic (approved):

PR #393, opened 2026-09-11
fastrpc: validate DSP-supplied buffer lengths in unpack_in_bufs

### LINKED ISSUE #392: fastrpc: DSP-supplied buffer lengths are not validated before the sbuf cursor advances (SIGSEGV; CVE-2020-11206 pattern still present)
### FastRPC version
1bd8835b65620a898e0d17dda2834071040824bd (vulnerable code is byte-identical on current `development` HEAD d24751965)

### DSP firmware version
ADSP.VT.5.4-00283-NICOBAR_1.0 (build 2019-09-04; strings dspso.bin, stock V12.0.26.0 vendor image)

### SoC and device name
Qualcomm SM6125 (Snapdragon 665) — Xiaomi Mi A3 (laurel_sprout), running mainline Linux (NixOS)

### Kernel version
7.3.0-rc1 (mainline)

### What happened?
The ADSP default listener (`adsprpcd audiopd`) died with SIGSEGV in
`__memcpy_generic` called from `listener()` (libadsprpc.so+0x32c1c).
Forensics on the retained core: `si_addr = input_buffer_base + 0xc8e8e6e7`
against a 4 KiB input buffer; the last decoded length word on the stack was
`0x6474735f` (~1.57 GiB).

Root cause, reproduced on host and on the device itself with a minimal
gate that drives the real `unpack_in_bufs()`:

1. `unpack_in_bufs()` (inc/listener_buf.h) advances the sbuf cursor by a
   DSP-supplied 4-byte length **without validating it**. `pra[].buf.pv` is
   only assigned when the original `(int)len <= sbuf_left(buf)` comparison
   happens to hold; on mismatch pv is left stale/NULL while the cursor
   still escapes the buffer.
2. `sbuf_left()` (inc/sbuf.h) subtracts two `uintptr_t` into an `int`.
   With the escaped cursor the true remainder is -3370702567, which
   overflows int32 and wraps to +924264729, so the caller's bounds check
   passes and `sbuf_read()`'s memmove dereferences the wild cursor.

Gate results (lengths: 0x6474735f, remaining+1, 0x7fffffff, 0x80000000,
0xffffffff):

| headers | result |
|---|---|
| original, host x86-64 | SIGSEGV |
| original, device aarch64 (this phone) | SIGSEGV |
| AOSP `android-17.0.0_r1`, host | SIGSEGV |
| patched (see linked PR), device aarch64 | all contained, exit 0 |

This is the CVE-2020-11206 pattern ("buffer overflow in Fastrpc while
handling received parameters due to lack of validation on input
parameters"); SM6125 is in the affected list. The November 2020 fix shipped
only as a closed-source component, so this open-source tree never received
it — the identical unchecked pattern is still present on `development` HEAD
and in AOSP `platform/external/fastrpc` through android-17.0.0_r1.

Patch proposal linked: validates the length in unsigned arithmetic against
the true unsigned remainder before the cursor advances, and computes
`sbuf_left` unsigned with clamping. Patched fastrpc userspace has been
running on the device as the live audiopd listener since deployment:
Type=notify readiness reached, no SEGV, normal operation.

angelwzr:
Fixes #392.

## Problem

`inc/listener_buf.h` `unpack_in_bufs()` advances the sbuf cursor by a
DSP-supplied 4-byte length without validating it against the remaining
buffer space. `pra[ii].buf.pv` is only assigned when the original
`(int)len <= sbuf_left(buf)` comparison holds, so a hostile length or an
upstream parser desync leaves pv stale/NULL while the cursor escapes the
buffer.

`inc/sbuf.h` `sbuf_left()` subtracts two `uintptr_t` into an `int`. When
the cursor has escaped, the true remainder (observed in the field:
-3370702567) overflows int32 and wraps large-positive, so the caller's
bounds check passes and `sbuf_read()`'s memmove dereferences the wild
cursor.

## Field evidence

Observed on an SM6125 device (Xiaomi Mi A3 / laurel_sprout, mainline
Linux, this tree at 1bd8835b): the ADSP default listener died with
SIGSEGV in `__memcpy_generic` called from `listener()` at
libadsprpc.so+0x32c1c, `si_addr = buffer_base + 0xc8e8e6e7`, after a
decoded length of `0x6474735f`. SM6125 is in the CVE-2020-11206 affected
list; the closed-source fix from the November 2020 bulletin never landed
in this tree, and the identical unchecked pattern is still present in
AOSP `platform/external/fastrpc` through `android-17.0.0_r1` (the crash
reproduces against those headers too).

## Fix

- `unpack_in_bufs`: validate len in unsigned arithmetic against the true
  unsigned remainder BEFORE advancing the cursor; on mismatch return
  `AEE_EBADPARM` with the buffer untouched.
- caller (`listener_android.c`): check the return and reject the whole
  invocation (`goto invoke`) instead of silently bypassing the buffers.
- `sbuf_left`: compute the difference in `uintptr_t` (unsigned wrap is
  well-defined) and clamp to `INT_MAX` instead of truncating through int;
  doc comment updated to the post-fix semantics (overflow returns 0).

## Verification

Gate (`unpack_in_bufs` driven with lengths 0x6474735f [field-captured],
remaining+1, 0x7fffffff, 0x80000000, 0xffffffff):

| headers | result |
|---|---|
| original, host x86-64 | SIGSEGV |
| original, SM6125 device (aarch64) | SIGSEGV |
| AOSP android-17.0.0_r1, host | SIGSEGV |
| **this patch, host** | **all contained, exit 0** |
| **this patch, SM6125 device** | **all contained, exit 0** |

The patched fastrpc userspace is deployed as the live audiopd listener on
the device: Type=notify readiness reached, no SEGV, normal operation.


inc/sbuf.h
ekanshibu: can you please modify the comment here? Overflow now returns 0, not a negative value.

angelwzr:
Friendly ping — this PR has been idle since Sep 11 with the required QSWAT `development-check` failing: the bot wants a `CRs-Fixed:` number (or a linked issue associated with a Change Request), and #392 doesn't have one associated yet.

@quic-vkatoch @ekanshibu — could either of you associate issue #392 with a CR (or hand me a number)? I'll add the `CRs-Fixed:` tag to the description as soon as one exists.

Context for review: the fix is small (+18/−6 across `inc/listener_buf.h` and `inc/sbuf.h`), DCO-clean, and based on the current `development` HEAD (`d2475196`). The SIGSEGV reproduction gate (host x86-64, device aarch64, AOSP headers) and the on-device audiopd verification are in the description — happy to make any changes needed.

inc/listener_buf.h
```diff
@@ -40,10 +40,19 @@ static __inline void unpack_in_bufs(struct sbuf* buf, remote_arg* pra, int nBufs
...
+         /* CVE-2020-11206 hardening: the length is DSP-supplied and must
+          * be validated in unsigned arithmetic BEFORE the cursor advances.
+          * The original code compared (int)len <= sbuf_left(buf), which a
+          * hostile length or an int-wrap in the remainder could defeat; the
+          * unsigned check below rejects the buffer untouched on mismatch. */
+         if (buf->bufCur > buf->bufEnd)
+            return;
+         remaining = (uintptr_t)(buf->bufEnd - buf->bufCur);
+         if ((uintptr_t)pra[ii].buf.nLen > remaining)
+            return; /* malformed: reject the whole buffer, do not advance */
```
ekanshibu: as this function is not returning any valid failure, the invalid scenarios might be bypassed, would you consider returning error code from here and checking the save at the caller side?

ekanshibu:
Please ignore the development-check failure.

angelwzr:
Force-pushed 16f30ba — both review comments addressed:

1. `inc/sbuf.h` — doc comment corrected: `sbuf_left()` returns 0 on overflow/exhaustion, never negative.
2. `inc/listener_buf.h` — `unpack_in_bufs()` now returns `int`. Escaped cursor or `len > remaining` returns `AEE_EBADPARM` with the buffer and cursor untouched; `listener_android.c` checks it and rejects the whole invocation via the existing `goto invoke` path, so the DSP gets an error response instead of a dead listener.

Gate re-run against the new signature: the field-captured `0x6474735f`, `0x7fffffff`, `0x80000000`, `0xffffffff`, remaining+1 and 0x100 all return `AEE_EBADPARM` (no SEGV); a valid length is still accepted.

And thanks — development-check waiver noted.

inc/listener_buf.h
```diff
@@ -32,21 +33,32 @@ static __inline void pack_out_lens(struct sbuf* buf, remote_arg* pra, int nBufs)
    }
 }
 
-static __inline void unpack_in_bufs(struct sbuf* buf, remote_arg* pra, int nBufs) {
-   int ii;
+static __inline int unpack_in_bufs(struct sbuf* buf, remote_arg* pra, int nBufs) {
+   int ii, nErr = AEE_SUCCESS;
```
ekanshibu: nErr is unused, you can just `return AEE_SUCCESS` in success case

angelwzr:
Pushed 9f4c373: `nErr` removed — `unpack_in_bufs()` returns `AEE_SUCCESS` directly on success; `AEE_EBADPARM` paths untouched. (The `nErr` in `unpack_out_bufs()` is pre-existing upstream code used by `VERIFY`/bail and is left alone.)

Gate re-run against this revision: field-captured `0x6474735f`, remaining+1, `0x7fffffff`, `0x80000000`, `0xffffffff` — all contained, exit 0.

ekanshibu (approved):

angelwzr:
Hi @ekanshibu, thanks for the review and for approving.

Quick question about CI — the `development-check` job has been failing on this PR since I opened it, while everything else passes (build, ABI, CodeQL, DCO, etc.). All I can see is that the log lives behind an internal link, so I can't tell what's actually failing.

Is this something I should be fixing on my side, or is it out of my hands? Happy to take a look if you can share what the job reports.

